PR #3, opened 2016-07-15
Simpler Nav

ericfranz:
Split the dropdowns into multiple partials and add two partials, simple_nav and nav for the nav that only contains Files & Clusters for Open OnDemand and the nav that contains everything for OSC.

This is to postpone making the nav more flexible until requirements and a better solution/design are decided on.


ericfranz:
@nickjer this is a temporary solution for https://github.com/AweSim-OSC/osc-ondemand/issues/116 till there is more time to implement a more proper solution (started in the menu branch)

This successfully lets Open OnDemand users that install in the next week or two to display Files, Clusters, or both depending on what they have deployed.

PR #7, opened 2016-08-11
Minor documentation improvements

treydock:
May be worth replacing initial scl command with something like this for all commands that require SCL.

```
scl enable nodejs010 rh-ruby22 -- bin/bundle install --path vendor/bundle
```

Using `scl enable nodejs010 rh-ruby22 bash` modifies root's shell in a such a way where an admin must be careful to exit that shell once done with the install.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,12 +9,12 @@ This Rails app doesn't use a database.
 1\. First enable software collections, checkout the code, and install the dependencies:
 
 ```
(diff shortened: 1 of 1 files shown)


ericfranz:
Sounds good. I'll merge your changes then update the README to prefix `scl enable nodejs010 rh-ruby22 --` in front of the commands

PR #13, opened 2016-09-15
App sharing (partial)

ericfranz:


ericfranz:
* Most of the code here is about modifying `app/apps/` model objects to deal with `OodApp` and `Router` classes which combine a path and a url, so that manifest information can be read from each app - without loading the app itself.

* A cleaner pull request would omit these files, which may eventually be removed or replaced:
  * app/assets/stylesheets/apps.scss
  * app/models/app_view.rb
  * app/models/user_with_shared_apps.rb
  * app/views/apps/_app.html.erb
  * app/views/apps/index.html.erb
  * test/controllers/apps_controller_test.rb

* `AppsController#show` was modified to use this
* `test/models/app_router_test.rb` needs more comments

Some are unrelated bugfixes:

* test/controllers/errors_controller_test.rb
* nav bar refactor
  * app/views/layouts/nav/_jobs_dropdown.html.erb
  * app/views/layouts/nav/_develop_dropdown.html.erb
  * etc.
* errors refactor including removing app/views/layouts/errors.html.erb

PR #18, opened 2016-09-29
Updated to Rails 4.2

nickjer:
Updated the code to use Rails 4.2.


config/application.rb
```diff
@@ -24,6 +28,5 @@ class Application < Rails::Application
     # The default locale is :en and all translations from config/locales/*.rb,yml are auto loaded.
     # config.i18n.load_path += Dir[Rails.root.join('my', 'locales', '*.{rb,yml}').to_s]
     # config.i18n.default_locale = :de
-    config.exceptions_app = self.routes
```
ericfranz: Should add this line back

```
config.exceptions_app = self.routes
```

Unless this is handled somewhere else.

nickjer: Done.

PR #20, opened 2016-09-29
App Sharing Sections / App Groups

ericfranz:
Fix apps/index/usr to list all of the app groups for all of the shared apps shared by each user


ericfranz:
@nickjer please review

PR #21, opened 2016-09-30
[WIP] Allow product (app) management by user

nickjer:
Treat products (apps) as a resource => `create, show, update, destroy`


ericfranz:
Lets not version the dotgit directory. In fact, should we even version a new app directory?

Part of the dashboard build directions could include:

1. running the rails application template to generate a new app into vendor/new_app
2. running bundle install (as currently the build directions specify)
3. cd-ing into the new_app directory and running git init + git add . + git commit (or whatever was done here) if we want to do this instead of running git init when we create a new app

nickjer:
1. Where are we getting `rails` from? If I download it on the fly will add an extra 2 minutes.
2. This is the most time-consuming part of the process.
3. I am fine with this.

The goal of this is to make new app creation as simple and seamless as possible. Having the user copy and paste a large set of instructions into a completely different app can be cumbersome and annoying.

ericfranz:
Pulled in because it was safe to add and now the dashboard can be fixed with links to the proper pages for "dev apps" and "your apps".

A separate pull request for future development can be opened.

PR #27, opened 2016-10-12
App Groups

ericfranz:
Fixes https://github.com/AweSim-OSC/osc-ondemand/issues/138 and https://github.com/AweSim-OSC/osc-ondemand/issues/137

Manifest lets you specify `group`.
Dashboard uses this to group apps by title.

4 questions:

1. should we restrict which app owners can take advantage of this feature?
2. with this approach, if a wiag shared app we deploy doesn't specify group, the default title will be Basil Gohar - is this bad? should we do this differently i.e. a dot file under ~/awesim/share to specify the default group title to use instead of the geocos - if none is specified
3. in parens, instead of just displaying the username, maybe we should include the geocos too IF the title is replaced by a custom title
4. `group` is ugly because it can be confused with a user group - but this is an "app group". Is there something better?

### DIFF
--- a/.env.test
+++ b/.env.test
@@ -1,0 +1,1 @@
+OOD_DATAROOT=$PWD/data
--- a/app/apps/dev_router.rb
+++ b/app/apps/dev_router.rb
@@ -1,9 +1,10 @@
 class DevRouter
-  attr_reader :name, :owner
+  attr_reader :name, :owner, :caption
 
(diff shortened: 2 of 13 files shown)


ericfranz:
All concerns addressed. We group now using `category`, display a `caption` above each description field. Default `category` is the owner of the app, but the category can be changed using the manifest.

ericfranz:
@nickjer please review

PR #28, opened 2016-10-12
Extend app sharing

nickjer:
Extended with a few more features & fixes:

  - instead of deleting app which leads to issues we just move the app to `.trash` with a timestamp (this fixes those nasty `.nfs*` files being left behind)
  - maybe add an "empty my trash" feature?
  - now we can get a list of active users running the app currently!!!1!1!
  - more to come...

### DIFF
--- a/app/models/product.rb
+++ b/app/models/product.rb
@@ -139,8 +139,8 @@ class Product
   end
 
   def destroy
-    trash_path.mkpath
-    FileUtils.mv router.path, trash_path.join("#{Time.now.localtime.strftime('%Y%m%dT%H%M%S')}_#{name}")
+    self.class.trash_path.mkpath
+    FileUtils.mv router.path, self.class.trash_path.join("#{Time.now.localtime.strftime('%Y%m%dT%H%M%S')}_#{name}")
   end
 
   def permissions?
--- a/app/models/product.rb
+++ b/app/models/product.rb
@@ -81,6 +81,14 @@ class Product
     rescue Errno::EACCES
       false
     end
+
+    def trash_path
+      router.base_path.join(".trash")
+    end
+
+    def trash_contents
+      trash_path.directory? ? trash_path.children : []
+    end
   end
 
   def app
@@ -131,7 +139,6 @@ class Product
   end
 
   def destroy
-    trash_path = self.class.router.base_path.join(".trash")
     trash_path.mkpath
     FileUtils.mv router.path, trash_path.join("#{Time.now.localtime.strftime('%Y%m%dT%H%M%S')}_#{name}")
   end
--- a/app/models/product.rb
+++ b/app/models/product.rb
@@ -131,8 +131,9 @@ class Product
   end
 
   def destroy
-    `kill -9 $(lsof -t #{router.path}) && sleep 1`
-    FileUtils.rm_rf(router.path)
+    trash_path = self.class.router.base_path.join(".trash")
+    trash_path.mkpath
+    FileUtils.mv router.path, trash_path.join("#{Time.now.localtime.strftime('%Y%m%dT%H%M%S')}_#{name}")
   end
 
   def permissions?
--- a/app/models/product.rb
+++ b/app/models/product.rb
@@ -156,9 +156,8 @@ class Product
     permissions(:group)
   end
 
-  def running_users
-    out = `ps -o uid= -p $(pgrep -f '^Passenger .*#{Regexp.quote(router.path.realdirpath.to_s)}') 2> /dev/null | sort | uniq`
-    out.split.map(&:to_i).map do |id|
+  def active_users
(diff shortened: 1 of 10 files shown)


nickjer:
image

image


image

nickjer:
@ericfranz take a look

app/models/product.rb
```diff
@@ -155,6 +163,12 @@ def groups
     permissions(:group)
   end
 
+  def active_users
+    @active_users ||= `ps -o uid= -p $(pgrep -f '^Passenger .*#{Regexp.quote(router.path.realdirpath.to_s)}') 2> /dev/null | sort | uniq`.split.map(&:to_i).map do |id|
+      OodSupport::User.new id
```
ericfranz: This can crash if the result ever gives you an `id` that is not an actual user.

nickjer: That is true, but not sure how the `ps` command could give us a `uid` that isn't an actual user.

app/views/products/index.html.erb
```diff
@@ -1,12 +1,21 @@
...
 <h2 class="apps-section-header">
   <%= products_title(@type) %>
 </h2>
 
 <p>
   <%= link_to 'New Product', new_product_path(type: @type), class: 'btn btn-default' %>
   <span class="pull-right">
-    <%= link_to 'Launch Shell', OodAppkit.shell.url(path: @type == :dev ? DevRouter.base_path : UsrRouter.base_path.realpath).to_s, target: '_blank', class: 'btn btn-default' %>
-    <%= link_to 'Launch Files', OodAppkit.files.url(path: @type == :dev ? DevRouter.base_path : UsrRouter.base_path.realpath).to_s, target: '_blank', class: 'btn btn-default' %>
+    <%= link_to 'Launch Shell', OodAppkit.shell.url(host: ENV['OOD_DEV_SSH_HOST'] || 'default', path: product_class.router.base_path.realpath).to_s, target: '_blank', class: 'btn btn-default' %>
```
ericfranz: In several places we have this: `ENV['OOD_DEV_SSH_HOST'] || 'default'`. I wonder if we should require explicitly setting the dev host, since ending up at the `default` would be bad. Does it make more sense to disable the shell buttons altogether or hide the buttons if `OOD_DEV_SSH_HOST` is not set.

When deploying, I could set `OOD_DEV_SSH_HOST=default` if I really wanted to.

nickjer: My other option was to commit `.env` with `OOD_DEV_SSH_HOST=default`. And `.env.local` would have priority.

app/views/products/_product.html.erb
```diff
@@ -17,12 +20,27 @@
...
-    <%= link_to 'Shell', OodAppkit.shell.url(path: product.router.path).to_s, target: '_blank', class: 'btn btn-default btn-block' %>
+    <%= link_to 'Shell', OodAppkit.shell.url(host: ENV['OOD_DEV_SSH_HOST'] || 'default', path: product.router.path).to_s, target: '_blank', class: 'btn btn-default btn-block' %>
     <%= link_to 'Files', OodAppkit.files.url(path: product.router.path).to_s, target: '_blank', class: 'btn btn-default btn-block' %>
   </td>
   <td>
     <%= link_to 'Details', product_path(product.name, type: @type), class: 'btn btn-info btn-block' %>
     <%= link_to 'Edit', edit_product_path(product.name, type: @type), class: 'btn btn-default btn-block' %>
   </td>
-  <td><%= link_to 'Delete', product_path(product.name, type: @type), method: :delete, data: { confirm: 'Are you sure?' }, class: 'btn btn-danger pull-right' %></td>
+  <td>
```
ericfranz: I didn't check close, is this identical to the one in app/views/products/show.html.erb? Should we use a partial?

nickjer: It is identical up to the variable that is passed in. And I'd throw this into a helper if necessary. Although you could say the same for all the buttons then.

nickjer: None of our apps set buttons in helpers or partials, so didn't want to break convention.

ericfranz: I have a feeling this string/button is more likely to change than the others:

>Unexpected bad things will happen if you don\'t read this!</div><p>This will move the product to the trash. Any users currently running this app will experience breaking behavior. Any past or present users of this app will be unable to access their data for this app. Be sure you know what you are doing!

Or the title:

> Are you ABSOLUTELY sure?

Could be wrong. I think breaking convention here is okay with a helper. My guess is it would be something like 

```erb
<%= delete_product_btn(@product, @type) %>
```

But if you feel strongly against it we can keep it as it is.

app/views/products/index.html.erb
ericfranz: Lets keep it as it is.

Issue #19, opened 2016-09-29
Error when `motd` is is empty file.


brianmcmichael:
When installing ood on a new machine, if there is no `motd` an exception is thrown.

1.) Simple fix is to update to return 0 messages if the file is missing or corrupt.

2.) Next step would be to remove mention of the "Message of the Day" if the file is missing or corrupt.

3.) May want to consider converting to RSS or something since the `motd` is so highly dependent on a formatting structure provided to OSC client services. We should not be prescribing a MOTD format for other sites.

ericfranz:
@brianmcmichael you seem to have a handle on this problem; how about:

![screen shot 2016-11-02 at 10 54 03 am](https://cloud.githubusercontent.com/assets/512333/19933699/b6ab4800-a0ea-11e6-882d-1894a4eb58ca.png)

```html
<h3>Message of the Day</h3>
<hr>
<p class="lead">No messages available.</p>
```

I also like the idea of motd => RSS => HTML but I'd open a separate FR for that.

PR #64, opened 2016-11-02
show 'no messages available' when the motd file does not exist

brianmcmichael:
submitted in response to https://github.com/OSC/ood-dashboard/issues/19#issuecomment-257889404

I used the text suggested, but perhaps it could be generalized to avoid duplication with the text for a situation where there are no messages. Alternatively, the message and logic could be updated to more explicitly mention that a motd was not found.

TBH though, I don't know if this is an improvement over the existing behavior, where if the motd is missing it just doesn't show anything. Not sure.

screen shot 2016-11-02 at 3 51 28 pm

### DIFF
--- a/app/views/dashboard/_motd.html.erb
+++ b/app/views/dashboard/_motd.html.erb
@@ -1,18 +1,29 @@
 <h3>Message of the Day</h3>
 
-<% if motd.messages.each do |m| %>
+<% if motd.exist? %>
 
-  <hr>
-  <div class="motd">
-    <h4 class="motd_title"><%= m.date %> - <%= m.title %></h4>
-    <div class="motd_body"><p><%=raw markdown( m.body ) %></p></div>
-  </div>
+  <% if motd.messages.each do |m| %>
 
-<% end.empty? %>
+    <hr>
+    <div class="motd">
+      <h4 class="motd_title"><%= m.date %> - <%= m.title %></h4>
+      <div class="motd_body"><p><%=raw markdown( m.body ) %></p></div>
+    </div>
 
-  <hr>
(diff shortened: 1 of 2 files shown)


ericfranz:
Oh I see. That is a good point. Essentially, we have 2 cases:

1. the motd file exists, but no well-formatted messages are available (parsing error) - preexisting behavior: display nothing
2. the motd file does not exist - preexisting behavior: crashes

Is that correct?

ericfranz:
Actually, I'm confused. How does this change fix the crash?

brianmcmichael:
The error was actually "fixed" at https://github.com/OSC/ood-dashboard/commit/f6bbfd32bbb765173150892566340c35f5fb0b5e but it looks like it didn't get pulled in until after the issue was created.

Currently this bug does not exist in v1.4.0

Not pulling in this commit will result in whitespace below the logo. Which might be better behavior anyways. However, if a developer is trying to configure the motd it might be helpful to show that the file was not found.

brianmcmichael:
I'm double-checking, but I can not reproduce this error on webdev02 when I load it without a configured motd, or with a malformed motd path.

If the error is still present in the current version, it does not seem like the cause is the existing display logic and may be due to another set of circumstances.

ericfranz:
The crash occurs when the motd file exists and is specified but is an empty file.

brianmcmichael:
Ok, great, I'll look into it.

brianmcmichael:
This PR is probably no longer necessary due to commits merged at https://github.com/OSC/ood-dashboard/commit/201f498fd434c8e4bde5414b419f1fec777b9c4d

Recommend closing.

PR #66, opened 2016-11-03
fix bug where empty motd file causes exception

brianmcmichael:
Fixes https://github.com/OSC/ood-dashboard/issues/19

Throw and catch an ArgumentError when the motd parser can't get any sections.

I feel like more things can go wrong with that regex in the wild though, and we may want to consider another rescue block that catches any other types of exceptions from that block and moves on.

### DIFF
--- a/app/models/motd_file.rb
+++ b/app/models/motd_file.rb
@@ -47,10 +47,15 @@ class MotdFile
 
     # get array of sections which are delimited by a row of ******
     sections = f.split(/^\*+$/).map(&:strip).select { |x| ! x.empty?  }
+    raise ArgumentError if sections.empty?
     sections.map! { |s| MotdFile::Message.from(s) }.compact!.sort_by! {|s| s.date }.reverse!
   rescue Errno::ENOENT
     # The messages file does not exist on the system.
-    logger.warn "MOTD File is missing; it was expected at #{motd_system_file}"
+    Rails.logger.warn "MOTD File is missing; it was expected at #{motd_system_file}"
+    []
+  rescue ArgumentError
+    # The messages file exists on the system but it empty or improperly formatted.
+    Rails.logger.warn "MOTD File is empty or incorrectly formatted; update it at #{motd_system_file}"
     []
   end
 
--- a/app/models/motd_file.rb
+++ b/app/models/motd_file.rb
@@ -47,7 +47,7 @@ class MotdFile
 
     # get array of sections which are delimited by a row of ******
     sections = f.split(/^\*+$/).map(&:strip).select { |x| ! x.empty?  }
-    sections.map! { |s| MotdFile::Message.from(s) }.compact.sort_by {|s| s.date }.reverse
+    sections.map { |s| MotdFile::Message.from(s) }.compact.sort_by {|s| s.date }.reverse
   rescue Errno::ENOENT
     # The messages file does not exist on the system.
     Rails.logger.warn "MOTD File is missing; it was expected at #{motd_system_file}"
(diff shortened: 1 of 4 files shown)


ericfranz:
Actually, it is the use of `compact!` instead of `compact` that is the issue.

Notice:

```ruby
2.2.2 :013 > [].compact!
 => nil
2.2.2 :014 > [].reject!(&:nil?)
 => nil
2.2.2 :015 > [].compact
 => []
2.2.2 :016 > [].reject(&:nil?)
 => []
2.2.2 :017 > [nil, 1].compact
 => [1]
2.2.2 :018 > [nil, 1].reject(&:nil?)
 => [1]
2.2.2 :019 > [nil, 1].compact!
 => [1]
2.2.2 :020 > [nil, 1].reject!(&:nil?)
 => [1]
```

nickjer:
I find editing in-place leads to more issues than its extremely small optimization is worth.

brianmcmichael:
I'm with @nickjer These lines are pretty dense.

```
sections = f.split(/^\*+$/).map(&:strip).select { |x| ! x.empty?  } 
sections.map! { |s| MotdFile::Message.from(s) }.compact!.sort_by! {|s| s.date }.reverse! 
```

I can replace the `.compact` invocation, but there are 11 potential points of failure in those two lines.

ericfranz:
Remove the `!`

```ruby
sections = f.split(/^\*+$/).map(&:strip).select { |x| ! x.empty?  }
sections = sections.map { |s| MotdFile::Message.from(s) }.compact.sort_by {|s| s.date }.reverse
```

* if `split` returns an array of `[]` then `map` and `select`do nothing; `strip` on a string will always return a string (not nil)
* if sections is [] then map, compact, sort_by, reverse are all ignored

```
irb(main):004:0> [].compact.sort_by {|s| s.date }.reverse
=> []
irb(main):005:0> [].map { |s| MotdFile::Message.from(s) }.compact.sort_by {|s| s.date }.reverse
=> []
```

So its a bunch of functional transformations but except for the regex there is no mystery what each one does. So there really isn't a point of "failure" in that sense.

ericfranz:
Though you could probably change this to a single `reduce/inject` call if it would improve readability.

ericfranz:
For example, perhaps `MotdFile::Message.from(s)` actually calls `strip` on the message and returns nil if the resulting text is empty. Then you would just need:

```ruby
sections = f.split(/^\*+$/).map { |s| MotdFile::Message.from(s) }.compact.sort_by {|s| s.date }.reverse
```

I'm not aware of a sorted array class in Ruby (or a way to keep sorted) but that would be nice too. Though if you had Motd::Message implement https://ruby-doc.org/core-2.2.3/Comparable.html then you could implement the spaceship operator to have this sort using `sort_by {|s| s.date }.reverse` by default. Then you could get by with this:

```ruby
sections = f.split(/^\*+$/).map { |s| MotdFile::Message.from(s) }.compact.sort
```

Thats much better.

app/models/motd_file.rb
```diff
@@ -47,10 +47,10 @@ def messages
 
     # get array of sections which are delimited by a row of ******
     sections = f.split(/^\*+$/).map(&:strip).select { |x| ! x.empty?  }
-    sections.map! { |s| MotdFile::Message.from(s) }.compact!.sort_by! {|s| s.date }.reverse!
+    sections.map! { |s| MotdFile::Message.from(s) }.compact.sort_by {|s| s.date }.reverse
```
ericfranz: go ahead and replace `map!` with `map` as well

Issue #58, opened 2016-10-26
Build nav bar dynamically based on sys apps manifest attributes


ericfranz:
Sys apps deployed can specify the category, subcategory, icon, title, and description - everything that we need to build a navbar menu. Finishing #57 will let us do this.
1. in an initializer, specify the categories of apps we will display in the navbar (and in what order)
2. use SysRouter.base_path to get a list of all the sys apps, and filter out those in each category to build a navbar
3. if apps specify a subcategory, further group those when building that dropdown
4. use the description in the popover
5. show the Develop menu if ENV['APP_SHARING'] is present (like before)
6. decide what to do with the Apps link that is on the AweSim dashboard - do we change this to "Home" and always have it available? Do we get rid of this altogether and have the user click on the icon in the top left?
7. Still keep OSC specific flag for project space links for now (address in separate issue)
- 5-7 could be addressed in separate issue
- Consider using http://martinfowler.com/eaaCatalog/twoStepView.html with some of the Nav and Nav::Link objects if it makes sense.
- Update installation documentation and other apps manifests first

ericfranz:
https://github.com/OSC/ood-dashboard/pull/67 fixed it

PR #67, opened 2016-11-03
Generate nav bar dynamically from system installed apps

### LINKED ISSUE #25: Navbar changes when an error is displayed
Before error...

image

During error...

image

ericfranz:
This addresses #58 and should also fix #25 and #2 and #69


ericfranz:
Left to do:

- [ ] make dropdown list configurable https://github.com/OSC/ood-dashboard/blob/43c99050284e96966808e240060215cc9be45818/app/controllers/application_controller.rb#L22 - for AweSim apps, hide this menu
- [ ] add home page link
- [ ] get rid of OSC specific file paths
- [ ] hide ruby desktops from non ruby users (use group ownership via chgrp on app directory OR add app dependency on Ruby cluster to manifest)
- [ ] update documentation

Other options:

- [ ] add unit test for manifest
- [ ] maybe load SysRouter.apps one time in an initializer (or set @nav_groups one time) https://github.com/OSC/ood-dashboard/blob/43c99050284e96966808e240060215cc9be45818/app/controllers/application_controller.rb#L22

config/initializers/ood.rb
```diff
@@ -0,0 +1,9 @@
+Rails.application.configure do
+  config.x.ood.nav_categories = ENV['OOD_NAV_CATEGORIES'] ? ENV['OOD_NAV_CATEGORIES'].split(",") : ["Files", "Jobs", "Clusters", "Desktops"]
+
+  #TODO: determine how to configure this via ENV
+  projects = User.new.groups.map(&:name).grep(/^P./)
+  config.x.ood.files_other_paths = projects.map { |p| Pathname.new("/fs/project/#{p}") }
+  config.x.ood.files_other_paths << Pathname.new("/fs/scratch/#{User.new.name}")
+  config.x.ood.files_other_paths << Pathname.new("/fs/scratch/#{projects.first}")
```
ericfranz: Instead of doing this, we could specify colon delimited files paths in an `OOD_FILES_PATHS` env var and use bash:

```
OOD_FILES_PATHS=`groups | ruby -e 'puts ARGF.read.scan(/(P\S*)/).map { |p| "/fs/project/#{p}:/fs/scratch/#{p}"}.join(":")'`
OOD_FILES_PATHS=$OOD_FILES_PATHS:/fs/scratch/$USER
```

Not sure if there is a more clever or succinct way to do this in bash.

PR #76, opened 2016-11-09
[DO NOT MERGE] move nav Help dd config to initializer

ericfranz:
This is an example of moving the navbar Help dropdown configuration
to the initializer. Instead of specifying a fixed number of URLs that can be specified,
let the installer specify any number of links.

I actually don't like this: asking the sys admin to write code like this:

```ruby
NavConfig.help_dropdown_links << NavLink.new(icon: 'question-circle', text: 'Contact Support', url: "https://www.osc.edu/contact/supercomputing_support")
```

However, we could actually maintain the status quo with the ENV vars and still go this route by simply adding another initializer `ood_nav_urls.rb` and do something like this:

```ruby
NavConfig.help_dropdown_links << NavLink.new(icon: 'question-circle', text: 'Contact Support', url: ENV['OOD_DASHBOARD_SUPPORT_URL']) if ENV['OOD_DASHBOARD_SUPPORT_URL']
```

### DIFF
--- a/app/apps/nav_config.rb
+++ b/app/apps/nav_config.rb
@@ -1,6 +1,10 @@
 class NavConfig
   class << self
-    attr_accessor :categories
+    attr_accessor :categories, :help_dropdown_links, :user_dropdown_links
   end
   self.categories = ["Files", "Jobs", "Clusters", "Desktops"]
+
+  # array of NavLink objects
+  self.help_dropdown_links = []
+  self.user_dropdown_links = []
 end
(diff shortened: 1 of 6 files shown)


ericfranz:
Is this any better?

```ruby
NavConfig.help_dropdown_links << {text: 'Contact Support', icon: 'question-circle', url: "https://www.osc.edu/contact/supercomputing_support"}
```

ericfranz:
Not doing. Did https://github.com/OSC/ood-dashboard/pull/77 instead. Can revisit this later if we want something more flexible.

PR #94, opened 2016-12-13
Develop dropdown visibility

### LINKED ISSUE #74: Determine whether or not to show Develop dropdown in an initializer
Instead of based on OOD_APP_SHARING enabled, check in an initializer. This of course can be the default (`ENV['OOD_APP_SHARING'].present?`) but we can change this so that instead for awesim, for instance, we check if the user is in a specific group OR is setup for app sharing, etc.

ericfranz:
Fixes #74

### DIFF
--- a/app/apps/nav_config.rb
+++ b/app/apps/nav_config.rb
@@ -1,6 +1,8 @@
 class NavConfig
   class << self
     attr_accessor :categories
+    attr_accessor :show_develop_dropdown
   end
   self.categories = ["Files", "Jobs", "Clusters", "Desktops"]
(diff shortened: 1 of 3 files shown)


ericfranz:
Maybe `NavConfig.developer_mode` instead of `NavConfig.show_develop_dropdown` so we can use this in other places where it makes sense.

nickjer:
Although I also wouldn't be opposed to a separate object that describes Developer properties.

nickjer:
I also would enable `developer_mode` as you have done, if `ENV["OOD_APP_SHARING"].present?`

PR #95, opened 2016-12-14
Product types

### LINKED ISSUE #93: Product model aware of type of app
Not only should the Product model be aware of the type of app:

  - passenger_ruby
  - passenger_nodejs
  - passenger_python
  - passenger_meteor
  - undetermined

The "details" view should conform to this app type (e.g., don't display `Bundle Install` button for a non-passenger_ruby app)

Can use the following for default determining app type:

https://www.phusionpassenger.com/library/config/nginx/reference/#passenger_startup_file

ericfranz:
Fixes #93

### DIFF
--- a/app/apps/ood_app.rb
+++ b/app/apps/ood_app.rb
@@ -84,6 +84,37 @@ class OodApp
     end
   end
 
+  def passenger_rack_app?
+    path.join("config.ru").file?
+  end
+
+  def passenger_nodejs_app?
+    path.join("app.js").file?
+  end
+
+  def passenger_python_app?
+    path.join("passenger_wsgi.py").file?
+  end
+
+  def passenger_meteor_app?
+    path.join(".meteor").exist?
+  end
+
+  def passenger_app?
+    passenger_rack_app? || passenger_nodejs_app? || passenger_python_app? || passenger_meteor_app?
+  end
+
+  def passenger_rails_app?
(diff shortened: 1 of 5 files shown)


ericfranz:
Probably the simplest solution. Doesn't prevent us from adding a "passenger_type" method that returns a symbol like :passenger_ruby which could be useful if we want to do a lot of small custom things based on the type and want to avoid lots of if/else blocks like `if @product.passenger_rails_app? else if @product.passenger_nodejs_app? etc` and want to get by with the passenger app type being the key of a hash. Also in the future we could make the commands a list on @product.

ericfranz:
Also doesn't account for the unlikely edge case that an app could be "recognized" as multiple types, i.e. a config.ru and an app.js in the same app root. I don't know how passenger handles this, and this edge case is probably not worth addressing.

ericfranz:
Final thought: we could probably refactor in the future to remove the need for the OodApp class and just use Product. But thats a separate discussion.

nickjer:
I think maybe using a helper method would be better than a partial for generating the CLI buttons.

Also, not sure if it is possible, but is there a way to have bootstrap collapse a `col` if it is empty? Otherwise the Terminal/Files buttons will appear in an odd location if the app isn't a Passenger app. A quick/dirty solution would be swapping the columns, so that those buttons appear on the right-hand side.

nickjer:
Assuming this all works, then I am fine with it.

Issue #87, opened 2016-11-30
When creating a new product, if no Git remote is provided, throw an error, instead of creating a new app


ericfranz:
When creating a "New Product", if I choose "Clone an existing repo" and forget to add the git remote for this repo and press Submit, but add a directory name, instead of throwing an error that git remote was not set, it creates a new app from the preinitialized app (as though I clicked Build a New App).

This was problematic for me because:

1. I clicked Clone Existing App because I wanted to clone an existing app, but when I accidentally pressed Enter after typing the Directory name in, it created a new app instead of cloning the app I wanted to clone
2. When I tried to clone the app again, the Directory name was now taken because it had created a new app in place where I wanted to clone an app to

Instead, require the Git remote to not be blank. 

![screen shot 2016-11-30 at 5 00 25 pm](https://cloud.githubusercontent.com/assets/512333/20773170/84f3b0aa-b71e-11e6-892d-bbfb6d448108.png)

Or it might be best just to instruct the user what to do:

![screen shot 2016-11-30 at 4 57 46 pm](https://cloud.githubusercontent.com/assets/512333/20773178/90a7d6ce-b71e-11e6-8921-f68f2ef70a2e.png)

Or offer a link to build a new app in the help text:

![screen shot 2016-11-30 at 4 59 15 pm](https://cloud.githubusercontent.com/assets/512333/20773181/940b9c1a-b71e-11e6-8c2a-4c1aa211fa58.png)

nickjer:
This is a bug. Since both forms for creating a new app post to the same URL, it determines that it is a git clone app if the git URL is provided otherwise it assumes it is a new app from the template.

nickjer:
Using a hidden field that describes what form (or what method to create the product from) sent in the new product request would be the simplest solution.

PR #97, opened 2016-12-15
Manifest fix

### LINKED ISSUE #96: Editing manifest through web interface deletes many attributes
Steps to reproduce:

1. Develop -> My Sandbox Apps -> New Product -> Clone Existing App
2. Clone the https://github.com/OSC/bc_osc_oakley_desktop app
3. Edit the manifest and change the description
4. Open a shell to the directory and do git diff

```diff
efranz@webtest04:~/awesim/dev/bc_osc_oakley_desktop (master)$ git diff
diff --git a/manifest.yml b/manifest.yml
index 833efd8..7113491 100644
--- a/manifest.yml
+++ b/manifest.yml
@@ -1,10 +1,5 @@
 ---
 name: Oakley Desktop
-description: |-
-  **Oakley Desktop** provides a remote desktop to one or more compute nodes on
-  the Oakley cluster at the Ohio Supercomputer Center (OSC).
-icon: fa://desktop
-category: Desktops
-subcategory: Interactive HPC
-role: ihpc
-url: /pun/dev/vncsim/%{app_token}/sessions
+description: "**Oakley Desktop** provides a remote desktop to one or more compute
+  nodes on\r\nthe Oakley cluster at the Ohio Supercomputer Center (OSC).\r\n\r\nWhat
+  happens?"
efranz@webtest04:~/awesim/dev/bc_osc_oakley_desktop (master)$ 
```

brianmcmichael:
Fixes #96 

Also added functionality that generates a new manifest if the old one is not valid.

### DIFF
--- a/app/apps/manifest.rb
+++ b/app/apps/manifest.rb
@@ -54,9 +54,24 @@ category: OSC
     {"name" => "", "description" => "", "category" => "", "subcategory" => "" , "icon" => "", "role" => "", "url" => ""}
   end
 
-  # Returns the contents of the Manifest object as YAML without the preceding `!ruby/object:Manifest`
+  # Creates a hash of the object's current state.
+  # @return [Hash] A hash representaion of the Manifest object.
+  def to_h
+    {
+        "name" => name,
+        "description" => description,
+        "category" => category,
+        "subcategory" => subcategory,
+        "icon" => icon,
+        "role" => role,
+        "url" => url
+    }
+  end
+
+  # Returns the contents of the object as a YAML string with the empty values removed.
+  # @return [String] The populated contents of the object as YAML string.
   def to_yaml
-    self.as_json.to_yaml
+    self.to_h.delete_if { |k, v| v.empty? }.to_yaml
   end
 
   def initialize(opts)
--- a/app/apps/manifest.rb
+++ b/app/apps/manifest.rb
@@ -142,8 +142,13 @@ category: OSC
     false
   end
 
-  def merge(hash)
+  def merge(opts)
+    raise InvalidContentError.new unless(opts && opts.is_a?(Hash))
 
+    @manifest_options.merge(opts.with_indifferent_access)
+
+    # merge with defaults, ignoring nil
+    # opts = defaults.merge(opts) { |key, oldval, newval| newval || oldval  }
     # TODO
   end
 
@@ -175,6 +180,10 @@ class InvalidManifest < Manifest
   def valid?
     false
   end
+
+  def save(path)
+    false
+  end
 end
 
 class MissingManifest < Manifest
@@ -183,6 +192,10 @@ class MissingManifest < Manifest
   end
 
   def exist?
+    false
+  end
+
+  def save(path)
     false
   end
 end
--- a/app/apps/manifest.rb
+++ b/app/apps/manifest.rb
@@ -71,44 +71,44 @@ category: OSC
 
   # The description of the application
   #
-  # @return [String] name as string
+  # @return [String] description as string
   def description
     @manifest_options[:description] || ""
   end
 
   # The icon used on the dashboard, optionally a Font Awesome tag
   #
-  # @return [String, nil]
+  # @return [String] icon as string
   def icon
-    @manifest_options[:icon]
+    @manifest_options[:icon] || ""
   end
 
   # Return the optional redirect URL string
   #
-  # @return [String, nil]
+  # @return [String] url as string
   def url
-    @manifest_options[:url]
+    @manifest_options[:url] || ""
   end
 
   # Return the app category
   #
-  # @return [String, nil]
+  # @return [String] category as string
   def category
-    @manifest_options[:category]
+    @manifest_options[:category] || ""
   end
 
   # Return the app subcategory
   #
-  # @return [String, nil]
+  # @return [String] subcategory as string
   def subcategory
-    @manifest_options[:subcategory]
+    @manifest_options[:subcategory] || ""
   end
 
   # Return the app role
   #
-  # @return [String, nil]
+  # @return [String] role as string
   def role
-    @manifest_options[:role]
+    @manifest_options[:role] || ""
   end
 
   # Manifest objects are valid
--- a/app/apps/manifest.rb
+++ b/app/apps/manifest.rb
@@ -134,9 +134,7 @@ category: OSC
   # @return [true, false] true if the file is saved successfully
   def save(path)
 
-    File.open(path.to_s, 'w') do |f|
-      f.write(self.to_yaml)
-    end
+    Pathname.new(path).write(self.to_yaml)
 
     true
   rescue
--- a/app/apps/manifest.rb
+++ b/app/apps/manifest.rb
@@ -150,14 +150,14 @@ category: OSC
   #
   # @return [Manifest] A new manifest with the updated options
   def merge(opts)
-    raise InvalidContentError.new unless(opts && (opts.is_a?(Hash) || opts.is_a?(Manifest)))
+    raise InvalidContentError.new unless(opts && opts.respond_to?(:to_h))
 
     Manifest.new(@manifest_options.merge(opts.to_h))
   end
 
   # Creates a hash of the object's current state.
   #
-  # @return [Hash] A hash representation of the Manifest object.
+  # @return [HashWithIndifferentAccess] A hash representation of the Manifest object.
   def to_h
     @manifest_options
   end
--- a/app/apps/manifest.rb
+++ b/app/apps/manifest.rb
@@ -144,7 +144,7 @@ category: OSC
   #
   # @return [String] The populated contents of the object as YAML string.
   def to_yaml
-    self.to_h.as_json.reject { |k, v| v.empty? }.to_yaml
+    self.to_h.as_json.to_yaml
   end
 
 end
--- a/app/apps/manifest.rb
+++ b/app/apps/manifest.rb
@@ -1,8 +1,6 @@
 require 'yaml'
 
 class Manifest
-  attr_accessor :name, :description
-  attr_reader :category, :subcategory, :icon, :role, :url
 
   class InvalidContentError < StandardError
     def initialize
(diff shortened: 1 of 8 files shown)


app/apps/manifest.rb
```diff
@@ -53,6 +54,11 @@ def defaults
     {"name" => "", "description" => "", "category" => "", "subcategory" => "" , "icon" => "", "role" => "", "url" => ""}
   end
 
+  # Returns the contents of the Manifest object as YAML without the preceding `!ruby/object:Manifest`
```
nickjer: In regards to a private convo with @ericfranz. `ActiveModel` does come with serialization features:

http://guides.rubyonrails.org/active_model_basics.html#serialization

app/models/product.rb
```diff
@@ -215,13 +215,15 @@ def stage
...
+      manifest.description = description unless description.blank?
       File.open(router.path.join('manifest.yml'), 'w') do |f|
-        f.write({
-          'name' => title,
-          'description' => description
-        }.to_yaml)
-      end if (!title.blank? || !description.blank?) || !router.path.join('manifest.yml').exist?
-      true
+        f.write(manifest.to_yaml)
+      end if router.path.join('manifest.yml').exist?
```
ericfranz: Don't force generating a manifest on `git clone`. The manifest is corrupt when clicking "Build New App" but we just need to fix the manifest in the starter app.

app/apps/manifest.rb
```diff
@@ -54,9 +54,24 @@ def defaults
...
+        "role" => role,
+        "url" => url
+    }
+  end
+
+  # Returns the contents of the object as a YAML string with the empty values removed.
+  # @return [String] The populated contents of the object as YAML string.
   def to_yaml
-    self.as_json.to_yaml
+    self.to_h.delete_if { |k, v| v.empty? }.to_yaml
```
ericfranz: Its a good practice to almost always use `reject` instead of `delete_if`; the difference is of course reject returns a new hash (or array) without modifying the original.

nickjer:
Fun tidbit, in Rails you can use `#compact` on a hash to clear out and `nil` objects:

```
def to_h
    {
        "name" => name,
        "description" => description,
        "category" => category,
        "subcategory" => subcategory,
        "icon" => icon,
        "role" => role,
        "url" => url
    }.compact
end
```

This way if one of those key-values is `nil`, it won't be defined in the yaml.

brianmcmichael:
We don't have any `nil` objects here because everything is being set to empty strings in `def defaults`

nickjer:
Ah, then it is strange that we are removing attributes from the YAML if it is blank even though that is a valid value.

brianmcmichael:
See the discussion at https://github.com/AweSim-OSC/rails-application-template/pull/31#commitcomment-20238264

nickjer:
I could see this being an issue if a developer wants a particular attribute to be blank, and we later on change the default value of that attribute from blank to some text.

Example:

Developer sets `description` to be blank. We later change default description to be `"description" => "Set me..."`. Then the developer will lose their current description and won't be able to set it blank again.

brianmcmichael:
Would we need to do that? 

Currently we handle these cases, if the description is blank we already provide text that says "Description should go here. Format using \*\*markdown\*\*." 

If the title is blank the folder name is used.

If the icon is blank the gear icon is used.

nickjer:
We don't do that for description. If the description is blank, then nothing shows up in the description for the Dashboard.

brianmcmichael:
Alternatively, this method does allow a developer to add a non-visible char and it will return a "blank" description in the YAML.

```
irb(main):003:0> m = Manifest.new({"name" => "brian", "description" => " "})
=> #<Manifest:0x00000006226068 @name="brian", @description=" ", @category="", @subcategory="", @icon="", @role="", @url="">
irb(main):004:0> m.to_yaml
=> "---\nname: brian\ndescription: \" \"\n"
irb(main):005:0> m = Manifest.new({"name" => "brian", "description" => "\n"})
=> #<Manifest:0x0000000624a440 @name="brian", @description="\n", @category="", @subcategory="", @icon="", @role="", @url="">
irb(main):006:0> m.to_yaml
=> "---\nname: brian\ndescription: |2+\n\n"
```

brianmcmichael:
What should `to_yaml` return if everything is empty?

Currently we get `"--- {}\n"`

Would it be better to return `nil` here?

ericfranz:
Ugly hack:

```ruby
self.to_h.reject { |k, v| k != "name" && k != "description" && v.empty? }
```

Then before rewriting a better solution, we should list all of the requirements, post it in an issue to rewrite or refactor this. Requirements including things like:

- when using in views, every value is guaranteed to be a string (it can't crash)
- values are properly sanitized as necessary (developers write their own manifests; we need to protect against them when loading it in our apps)
- don't spread knowledge of the manifest location in multiple places
- if manifest values are already set, preserve those, even if the values set are things like `""`

Since some cases, like an invalid or missing manifest being treated as failed validations on Product, this opens the possibility of doing something much simpler in Product itself.

Two related refactor issues: #100 and #102

app/models/product.rb
```diff
@@ -215,12 +215,19 @@ def stage
...
       File.open(router.path.join('manifest.yml'), 'w') do |f|
-        f.write({
-          'name' => title,
-          'description' => description
-        }.to_yaml)
+        f.write(manifest.to_yaml)
       end if (!title.blank? || !description.blank?) || !router.path.join('manifest.yml').exist?
+
       true
     end
```
ericfranz: After suggested changes the code would be:

```ruby

# OodApp#manifest_path should be created

Manifest.load(app.manifest_path).merge(name: title, description: description).save(manifest_path) if (!title.blank? || !description.blank?) || !app.manifest_path.exist?
```

If statement will go away after fixing https://github.com/OSC/ood-dashboard/issues/87

* save should throw exception and we should handle it here

app/models/product.rb
```diff
@@ -214,13 +214,17 @@ def stage
...
-      File.open(router.path.join('manifest.yml'), 'w') do |f|
-        f.write({
-          'name' => title,
-          'description' => description
-        }.to_yaml)
-      end if (!title.blank? || !description.blank?) || !router.path.join('manifest.yml').exist?
+      manifest = Manifest.load( router.path.join('manifest.yml') )
+
+      new_attributes = { name: title, description: description }
+      manifest = manifest.valid? ? manifest.merge(new_attributes) : Manifest.new(new_attributes)
```
ericfranz: `Manifest#merge` should return  new Manifest then you can do 

```ruby
manifest = manifest.merge({ name: title, description: description })
```

test/models/manifest_test.rb
```diff
@@ -0,0 +1,86 @@
...
+  @@hash4 = { name: nil, description: nil, role: nil, url: "http://127.0.0.1" }
+
+  test "load a valid manifest" do
+    manifest = Manifest.load("test/fixtures/files/manifest_valid")
+    assert_instance_of Manifest, manifest, "Not a Manifest Object"
+  end
+
+  test "load an invalid manifest" do
+    manifest = Manifest.load("test/fixtures/files/manifest_invalid")
+    assert_instance_of InvalidManifest, manifest, "Not an InvalidManifest Object"
```
ericfranz: For the tests, don't `assert_instance_of InvalidManifest` instead use `assert (! manifest.valid?)` or `refute manifest.valid?`

ericfranz: In other words, outside users shouldn't need to know that "ManifestMissing and "InvalidManifest" classes actually exist. That is simply a Manifest implementation detail. The interface is `Manifest#valid?`

BTW I just realized that in accordance with this fact, we should probably move `attr_reader :exception` to the Manifest superclass. It can remain nil. That way the interface is the same. That is my bad.

app/apps/manifest.rb
```diff
@@ -49,36 +50,128 @@ def self.load_from_string(yaml)
...
+    false
+  end
+
+  # Merge the contents of a hash into this Manifest's options
+  #
+  # @param [Hash, Manifest] opts The options to update
+  #
+  # @return [Manifest] A new manifest with the updated options
+  def merge(opts)
+    raise InvalidContentError.new unless(opts && (opts.is_a?(Hash) || opts.is_a?(Manifest)))
```
nickjer: Instead of checking if it is a specific Class. You could just do:

```ruby
opts.respond_to?(:to_h)
```

app/apps/manifest.rb
```diff
@@ -49,36 +50,128 @@ def self.load_from_string(yaml)
...
+
+    Manifest.new(@manifest_options.merge(opts.to_h))
+  end
+
+  # Creates a hash of the object's current state.
+  #
+  # @return [Hash] A hash representation of the Manifest object.
+  def to_h
+    @manifest_options
```
nickjer: Why don't we call `#compact` here? And instead wait until the `def to_yaml` portion.

app/apps/manifest.rb
```diff
@@ -49,36 +50,128 @@ def self.load_from_string(yaml)
...
+  end
+
+  # Returns the contents of the object as a YAML string with the empty values removed.
+  #
+  # @return [String] The populated contents of the object as YAML string.
+  def to_yaml
+    self.to_h.compact.as_json.to_yaml
```
nickjer: Is the `as_json` still necessary? Not sure what it is used for here.

app/apps/manifest.rb
```diff
@@ -49,36 +50,128 @@ def self.load_from_string(yaml)
...
+  # @param [Hash] opts A hash of the options in the manifest
+  # @option opts [String] :name The name of the application
+  # @option opts [String] :description The description of the application
+  # @option opts [String] :category The category of the application
+  # @option opts [String] :subcategory The subcategory of the application
+  # @option opts [String] :icon The icon used on the dashboard, optionally a Font Awesome tag
+  # @option opts [String] :role Dashboard categorization
+  # @option opts [String] :url An optional redirect URL
   def initialize(opts)
     raise InvalidContentError.new unless(opts && opts.is_a?(Hash))
```
nickjer: If you just checked for `opts.respond_to?(:to_h)` then you could get away with initializing a manifest from another manifest (similar to `Pathname`).

```ruby
def initialize(opts)
  raise InvalidContentError.new unless opts.respond_to?(:to_h)
  @manifest_options = opts.to_h.with_indifferent_access
end
```

Then you can do things like....

```ruby
m = Manifest.new(name: "Name")
h = {name: "Name"}

my_manifest = Manifest.new(m)
my_manifest = Manifest.new(h)
```

app/apps/manifest.rb
```diff
@@ -49,36 +50,128 @@ def self.load_from_string(yaml)
...
+  # Save the current manifest to a path.
+  #
+  # @param [String, Pathname] path The full path of the file to be saved as string or Pathname object
+  #
+  # @return [true, false] true if the file is saved successfully
+  def save(path)
+
+    File.open(path.to_s, 'w') do |f|
+      f.write(self.to_yaml)
+    end
```
ericfranz: You can probably get away with:

```ruby
Pathname.new(path).write(self.to_yaml)
```

app/models/product.rb
```diff
@@ -214,13 +214,16 @@ def stage
...
-        f.write({
-          'name' => title,
-          'description' => description
-        }.to_yaml)
-      end if (!title.blank? || !description.blank?) || !router.path.join('manifest.yml').exist?
+      manifest = Manifest.load( router.path.join('manifest.yml') )
+
+      manifest = manifest.merge({ name: title, description: description })
+
+      manifest.save(router.path.join('manifest.yml')) if (!title.blank? || !description.blank?) || !router.path.join('manifest.yml').exist?
```
ericfranz: `router.path.join('manifest.yml')` is duplicated here: https://github.com/OSC/ood-dashboard/blob/ea579eccff193a8b45f6e6d98b1aa5bcfdf385f2/app/apps/ood_app.rb#L139-L143

```ruby
  def load_manifest
    default = path.join("manifest.yml")
    alt = path.dirname.join("#{path.basename}.yml")
    alt.exist? ? Manifest.load(alt) : Manifest.load(default)
  end
```

Should add a public method OodApp#manifest_path and then update OodApp#manifest to look like this:

```diff

  def manifest
-    @manifest ||= load_manifest
+    @manifest ||= Manifest.load(manifest_path)
  end
```

The whole alt manifest thing in `OodApp#load_manifest` can go away. That is an artifact from the old dashboard.

ericfranz:
manifest_path should be used in Product#write_manifest

PR #105, opened 2016-12-19
add error message when git commit fails

### LINKED ISSUE #86: When creating a new product, if cloning fails, display the command run and the full error output to the user
Copied and reformatted from https://github.com/AweSim-OSC/awesim-dev-dashboard/issues/69

When creating a new "Product" for sandbox or shared apps, we fill out a form that includes specifying a Git remote. If the corresponding git clone command fails, the user only sees "was unable to be cloned" without seeing the actual command that ran or the actual error that resulted. Here is an example, where an SSH key was missing:

Here is another example:

1. Go to My Sandbox Apps
2. Click "New Product"
3. Click "Clone existing app"
4. Type in a git remote that does not exist (i.e. `git@github.com:AweSim-OSC/weld_predictorrrrrrrrrr.git`)
5. Click Submit

screen shot 2016-11-30 at 4 41 47 pm

Instead, add a colon after, and display both the command that was run and the output and error below the command:

752743da-a102-11e6-9f04-4f4b50e1f133

brianmcmichael:
Mostly Fixes #86 

The only issue is that the Open3 command doesn't maintain the command, so I only show the stdout and stderr here. If we want to add the shell command string I'll need to basically build one out of the Open3 commands at https://github.com/OSC/ood-dashboard/compare/cloneerror?expand=1#diff-7425083b688c32187c716cc5c53c664fR247

If you want me to go ahead with that let me know. Here's where we're at now:

cloneerror

### DIFF
--- a/app/models/product.rb
+++ b/app/models/product.rb
@@ -246,7 +246,8 @@ class Product
     def clone_git_repo(target)
       o, s = Open3.capture2e({"HOME" => ""}, "git", "clone", git_remote, target.to_s)
       unless s.success?
-        errors.add(:git_remote, "was unable to be cloned")
+        errors.add(:git_remote, "was unable to be cloned:")
+        errors.add(:git_remote_error, o)
         Rails.logger.error(o)
         return false
       end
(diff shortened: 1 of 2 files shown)


ericfranz:
I like it.

Think you might be able to do 

```ruby
<%= content_tag :pre, @product.errors[:git_remote_error].first.html_safe, class: "text-danger" if @product.errors[:git_remote_error].any? %>
```

brianmcmichael:
Updated to use the content_tag.

PR #109, opened 2017-01-03
Git require

brianmcmichael:
Fixes https://github.com/OSC/ood-dashboard/issues/87

Manual tests work to spec. Not sure if modifying the existing validation would have adverse effects elsewhere, so some test coverage might be warranted.

The view has a bunch of modifications, but most of that is just indenting the nested if statement for better readability. Pay particular attention to the addition of the hidden form field that allows us to pass validation.

### DIFF
--- a/app/controllers/products_controller.rb
+++ b/app/controllers/products_controller.rb
@@ -23,20 +23,25 @@ class ProductsController < ApplicationController
   # GET /products/new
   def new
     @type = params[:type].to_sym
+    # Only allow user to clone existing apps in shared environment
+    if @type == :usr
+      @new_method = 'git'
+      redirect_to new_from_git_remote_url
+    end
   end
 
   # GET /products/new_from_git_remote
   def new_from_git_remote
     @type = params[:type].to_sym
     @product = Product.build(type: @type)
-    @new_method = 'git' if @type == :usr
+    @new_method = @type == :usr ? 'git' : params[:new_method]
   end
 
   # GET /products/new_from_rails_template
   def new_from_rails_template
     @type = params[:type].to_sym
     @product = Product.build(type: @type)
-    @new_method = 'template' if @type == :usr
+    @new_method = @type == :usr ? 'git' : params[:new_method]
   end
 
   # GET /products/1/edit
@@ -50,7 +55,7 @@ class ProductsController < ApplicationController
   def create_from_git_remote
     @type = params[:type].to_sym
     @product = Product.build(product_params.merge(type: @type))
-    @new_method = 'git' if @type == :usr
+    @new_method = @type == :usr ? 'git' : params[:new_method]
 
     respond_to do |format|
       if @product.create_from_git_remote
@@ -68,7 +73,7 @@ class ProductsController < ApplicationController
   def create_from_rails_template
     @type = params[:type].to_sym
     @product = Product.build(product_params.merge(type: @type))
-    @new_method = 'template' if @type == :usr
+    @new_method = @type == :usr ? 'git' : params[:new_method]
 
     respond_to do |format|
       if @product.create_from_rails_template
--- a/app/controllers/products_controller.rb
+++ b/app/controllers/products_controller.rb
@@ -41,7 +41,7 @@ class ProductsController < ApplicationController
     @new_method = @type == :usr ? 'git' : params[:new_method]
 
     respond_to do |format|
-      if @product.save
+      if @product.save(context: @product.create_context)
         format.html { redirect_to product_url(@product.name, type: @type), notice: 'Product was successfully created.' }
         format.json { render :show, status: :created, location: @product }
       else
--- a/app/controllers/products_controller.rb
+++ b/app/controllers/products_controller.rb
@@ -23,6 +23,18 @@ class ProductsController < ApplicationController
   # GET /products/new
   def new
     @type = params[:type].to_sym
+  end
+
+  # GET /products/new_from_git_remote
+  def new_from_git_remote
+    @type = params[:type].to_sym
+    @product = Product.build(type: @type)
+    @new_method = @type == :usr ? 'git' : params[:new_method]
+  end
+
+  # GET /products/new_from_rails_template
+  def new_from_rails_template
+    @type = params[:type].to_sym
     @product = Product.build(type: @type)
     @new_method = @type == :usr ? 'git' : params[:new_method]
   end
@@ -42,6 +54,42 @@ class ProductsController < ApplicationController
 
     respond_to do |format|
       if @product.save(context: params[:create_context])
+        format.html { redirect_to product_url(@product.name, type: @type), notice: 'Product was successfully created.' }
+        format.json { render :show, status: :created, location: @product }
+      else
+        format.html { render :new }
+        format.json { render json: @product.errors, status: :unprocessable_entity }
+      end
+    end
+  end
+
+  # POST /products
+  # POST /products.json
+  def create_from_git_remote
+    @type = params[:type].to_sym
+    @product = Product.build(product_params.merge(type: @type))
+    @new_method = @type == :usr ? 'git' : params[:new_method]
+
+    respond_to do |format|
+      if @product.create_from_git_remote
+        format.html { redirect_to product_url(@product.name, type: @type), notice: 'Product was successfully created.' }
+        format.json { render :show, status: :created, location: @product }
+      else
+        format.html { render :new }
+        format.json { render json: @product.errors, status: :unprocessable_entity }
+      end
+    end
+  end
+
+  # POST /products
+  # POST /products.json
+  def create_from_rails_template
+    @type = params[:type].to_sym
+    @product = Product.build(product_params.merge(type: @type))
+    @new_method = @type == :usr ? 'git' : params[:new_method]
+
+    respond_to do |format|
+      if @product.create_from_rails_template
         format.html { redirect_to product_url(@product.name, type: @type), notice: 'Product was successfully created.' }
         format.json { render :show, status: :created, location: @product }
       else
--- a/app/controllers/products_controller.rb
+++ b/app/controllers/products_controller.rb
@@ -41,7 +41,7 @@ class ProductsController < ApplicationController
     @new_method = @type == :usr ? 'git' : params[:new_method]
 
     respond_to do |format|
-      if @product.save(context: @product.create_context)
+      if @product.save(context: params[:create_context])
         format.html { redirect_to product_url(@product.name, type: @type), notice: 'Product was successfully created.' }
         format.json { render :show, status: :created, location: @product }
       else
--- a/app/controllers/products_controller.rb
+++ b/app/controllers/products_controller.rb
@@ -45,26 +45,8 @@ class ProductsController < ApplicationController
     @product = Product.find(@type, params[:name])
   end
 
-  # POST /products
-  # POST /products.json
-  def create
-    @type = params[:type].to_sym
-    @product = Product.build(product_params.merge(type: @type))
-    @new_method = @type == :usr ? 'git' : params[:new_method]
-
-    respond_to do |format|
-      if @product.save(context: params[:create_context])
-        format.html { redirect_to product_url(@product.name, type: @type), notice: 'Product was successfully created.' }
-        format.json { render :show, status: :created, location: @product }
-      else
-        format.html { render :new }
-        format.json { render json: @product.errors, status: :unprocessable_entity }
-      end
-    end
-  end
-
-  # POST /products
-  # POST /products.json
+  # POST /create_from_git_remote
+  # POST /create_from_git_remote.json
   def create_from_git_remote
     @type = params[:type].to_sym
     @product = Product.build(product_params.merge(type: @type))
@@ -81,8 +63,8 @@ class ProductsController < ApplicationController
     end
   end
 
-  # POST /products
-  # POST /products.json
+  # POST /create_from_rails_template
+  # POST /create_from_rails_template.json
   def create_from_rails_template
     @type = params[:type].to_sym
     @product = Product.build(product_params.merge(type: @type))
--- a/app/controllers/products_controller.rb
+++ b/app/controllers/products_controller.rb
@@ -29,14 +29,14 @@ class ProductsController < ApplicationController
   def new_from_git_remote
     @type = params[:type].to_sym
     @product = Product.build(type: @type)
-    @new_method = @type == :usr ? 'git' : params[:new_method]
+    @new_method = 'git' if @type == :usr
   end
 
   # GET /products/new_from_rails_template
   def new_from_rails_template
     @type = params[:type].to_sym
     @product = Product.build(type: @type)
-    @new_method = @type == :usr ? 'git' : params[:new_method]
+    @new_method = 'template' if @type == :usr
   end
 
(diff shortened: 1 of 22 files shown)


ericfranz:
I'm okay with this for now. @nickjer?

ericfranz:
Actually I take that back. This makes me uncomfortable. Have to think about this for a bit.

brianmcmichael:
It's not perfect, but this is submitted as a fix for a bug. We've also talked about splitting this up into two separate actions. At that time, this could be developed into a cleaner solution.

nickjer:
Actually, that validation should probably be moved to `UsrProduct`.

Also, I am not sure we should be using the `git_remote` attribute as the flag to determine the action to create an app. I was leaning more towards some `action` or something similar that determines what action to take when creating an app in the `#save` method. So that `#save` would call something like:

- `#create_from_git`
- `#create_from_rails_template`
- ...

from within it.

ericfranz:
>I was leaning more towards some action or something similar that determines what action to take when creating an app in the #save method.

If we do that, it would probably be just as easy to have two different actions:

* `ProductsController#create_from_rails_template` - copy the template!
* `ProductsController#create` -  copy from git remote (using the web form)

> Actually, that validation should probably be moved to UsrProduct

Why is this only on UsrProduct?

Also, I think the issue is that we are using git remote in the webform to create an app from it, and we are using model validations to verify whether the webform was successful or not. So to not run this validation on the DevProduct as well would mean that we would need another method of validating this form. So now it gets a little ugly. Maybe a separate FormObject for validating the webform for creating a new product would make more sense.

nickjer:
> Why is this only on UsrProduct?

In the original design, the developer could create an app from a template or from git under the Sandbox (so we didn't force `git_remote` presence), but they could only create an app from a git repo for Production app (so we force `git_remote` presence).

Although this will change if we introduce different actions to create an app. For example this could now be on `Product`:

```ruby
validates :git_remote, presence: true, on: :create_from_git
```

Which is called in a `#create_from_git` method when it runs `self.valid?(:create_from_git)`.

>If we do that, it would probably be just as easy to have two different actions:
>
>* `ProductsController#create_from_rails_template` - copy the template!
>* `ProductsController#create` -  copy from git remote (using the web form)

I am fine with making it an action on the controller, but I believe the action `#create_from_rails_template` will also need a web form. That way the user can supply the directory name, app title, and app description.

Also, I wonder if we can come up with a bit more concise method names if we go this route. I just chose those longer names to explain my intentions in my thoughts.

ericfranz:
>Also, I wonder if we can come up with a bit more concise method names if we go this route. I just chose those longer names to explain my intentions in my thoughts.

The best names do just that: "explain my intentions in my thoughts"

I don't see any reason for more concise names. Those are great.

ericfranz:
For example https://signalvnoise.com/posts/3250-clarity-over-brevity-in-variable-and-method-names

ericfranz:
```ruby
```

ericfranz:
Okay, so you can scope validations based on custom contexts:

```ruby
validates :git_remote, presence: true, on: :create_from_git_remote
```

then do something like this:

```ruby
def create_from_rails_template
  if self.valid?(:create_from_rails_template)
    copy_rails_template_to(router.path) && write_manifest
  else
    false
  end
end

def create_from_git_remote
  if self.valid?(:create_from_git_remote)
    clone_git_repo_to(router.path) && write_manifest
  else
    false
  end
end
```

And this context could be set in the form via a hidden value:

```erb
<%= f.hidden_field :create_context, value: "create_from_git_remote" %>
```


Then, as Jeremy suggested, an option is doing something like this:

```ruby
def save(context:)
  case context.to_sym
    when :create_from_git_remote
      create_from_git_remote
    when :create_from_rails_template
      create_from_rails_template
    else
      raise "Context invalid"
  end
end
```



Having separate create methods like above removes the control flow statement that was previously present in the `stage` method, so it makes the code easier to read in the Product class.

There will still be control flow statements and some cleverness in `ProductsController#new` and `ProductsController#create` and in the views and using these hidden fields and get params to properly display the web forms for the different creation methods....unless you do this:

All the control flow statements can be removed by adding `ProductsController#new` and `ProductsController#new_from_git_remote` and `ProductsController#new_from_rails_template` and `ProductsController#create_from_rails_template` and `ProductsController#create_from_git_remote`, and ensuring that new only provides links to the `new_from_git_remote` and `new_from_rails_template` (with new views for new_from_rails_template and new_from_git_remote). Varying these will be slightly more complex because you are varying the form path instead of the form method (its a post to create_from_git_remote versus a post to create_from_rails_template). The tradeoffs: the code will be simpler, but there will be more of it.

As long as we have separate create methods in Product so we remove some control flow there, it will reduce some complexity and fix the bug, and I'm okay with that too.

nickjer:
You do not need `#write_manifest` inside of `#create_from_git_remote`. This will help clean up some of the `if` statements inside `#write_manifest`.

ericfranz:
Thats great.

Lets go with the full removal of control flow statements for now. We could always change routes in the future to specify a default value for `create_context` if we wanted to use a single controller action for these in the future. But from the views perspective they would be separate routes i.e. adding `ProductsController#new` and `ProductsController#new_from_git_remote` and `ProductsController#new_from_rails_template` and `ProductsController#create_from_rails_template` and `ProductsController#create_from_git_remote`.

brianmcmichael:
I've made an attempt at this refactoring but I can't find much documentation on the "context", so I gave it my best guess by adding an accessor to track. This works as expected but it's got some extra code and  I'm probably off-base in interpreting @ericfranz solution.

I've split the stage method into two distinct methods, but I didn't remove the existing `stage` method because it's called in the controller at https://github.com/OSC/ood-dashboard/blob/git_require/app/controllers/products_controller.rb#L8 I'll need to figure out how best to update this so that we can get rid of `Product#stage` since it's all duplicated elsewhere now.

Anyways, I could use a fresh set of eyes on this. This code feels wrong to me.

app/controllers/products_controller.rb
```diff
@@ -41,7 +41,7 @@ def create
     @new_method = @type == :usr ? 'git' : params[:new_method]
 
     respond_to do |format|
-      if @product.save
+      if @product.save(context: @product.create_context)
```
ericfranz: Instead of adding create_context as an attribute to Product, it makes more sense here that create_context would be just on the form and in params. So if you do this you might use `hidden_field` instead of `f.hidden_field` and then in params you could access it: `params[:create_context]` i.e. `@product.save(context: params[:create_context]`.

ericfranz: `Product#stage`, the method you split up, can be removed as it is never called. The code you call it is the use of `Product.stage`. They are two separate methods. They are both in the same file.

ericfranz:
The refactoring is a start, but right now the end goal is to see the removal of control flow statements by having separate "new" and "create". If you have separate controller methods, instead of calling `save` you can just call `create_from_rails_template` directly.

>Lets go with the full removal of control flow statements for now. We could always change routes in the future to specify a default value for create_context if we wanted to use a single controller action for these in the future. But from the views perspective they would be separate routes i.e. adding ProductsController#new and ProductsController#new_from_git_remote and ProductsController#new_from_rails_template and ProductsController#create_from_rails_template and ProductsController#create_from_git_remote.

ericfranz:
My guess is the reason you feel like the code feels wrong to you is because it hasn't reach the desired end point and thus we have this funny method:

```ruby
  def save(context:)
    case context.to_sym
      when :create_from_git_remote
        create_from_git_remote
      when :create_from_rails_template
        create_from_rails_template && write_manifest
      else
        raise "Invalid Context"
    end
  end
```

I'd move `write_manifest` call to inside `create_from_rails_template`. Then we are left with:

```ruby
  def save(context:)
    case context.to_sym
      when :create_from_git_remote
        create_from_git_remote
      when :create_from_rails_template
        create_from_rails_template
      else
        raise "Invalid Context"
    end
  end
```

At that point you see that we can get rid of this method completely if the controller calls `create_from_rails_template` or `create_from_git_remote` directly.

ericfranz:
Also might be able to remove some of the if statements in `write_manifest` that Jeremy mentioned

>You do not need #write_manifest inside of #create_from_git_remote. This will help clean up some of the if statements inside #write_manifest.

app/controllers/products_controller.rb
brianmcmichael: Thanks, that bootstrap_form helper was the cause of my grief.

brianmcmichael:
> Lets go with the full removal of control flow statements for now. We could always change routes in the future to specify a default value for create_context if we wanted to use a single controller action for these in the future. But from the views perspective they would be separate routes i.e. adding ProductsController#new and ProductsController#new_from_git_remote and ProductsController#new_from_rails_template and ProductsController#create_from_rails_template and ProductsController#create_from_git_remote.

Just to be clear here, I'll be adding four additional routes to avoid an `if` statement in the `save` method?

Also, it looks like by doing this we're still going to have a control statement, but it's going to be in the View instead of in the Model. I'm not sure that's better.

ericfranz:
Your latest commit removed write_manifest call from save ==> it needs to be added to `create_from_rails_template` directly.

> Just to be clear here, I'll be adding four additional routes to avoid an if statement in the save method?

If you add the routes, there will be no control statements in the view or the controller.

1. ProductsController#new will show two options with two buttons and two links => #new_from_git_remote and #new_from_rails_template
2. two separate actions handling these would have separate forms as a result

There would be more code, but all the code would look really simple. Theoretically.

brianmcmichael:
How  do we want to handle the "edit" action here? 

Currently "Edit" uses `_form.html.erb` which is no longer used by the create actions, however, it duplicates all of the code and is effectively the original form used by the "new" action.

Should I just go ahead and split `edit` up into additional actions/views as well?

I actually don't like the way edit is currently implemented anyways. For example, after creating a new app from a rails template, there is an error that there is no git repo assigned, however, when you click "edit", there is no field available to add a repo, since the logic blocks it.

It may be a better idea to keep `_form.html.erb` and generalize it so that all of the fields are available, then use this form for the edit action.

brianmcmichael:
Nevermind, there is currently no mechanism to add a git repo via the web interface. It looks like the edit form is only able to edit manifest entries. I may put the duplicated portions into partials to avoid so much code duplication between the `new` and `edit` actions.

brianmcmichael:
@ericfranz please review.

The new app creation process is now two distinct paths. I've used partials to cut down on some code duplication and cut out all of the extraneous code I could find.

app/views/products/new_from_git_remote.html.erb
```diff
@@ -0,0 +1,13 @@
+<%= render 'header' %>
+
+<%= bootstrap_form_for(@product, as: :product, url: create_from_git_remote_path(type: @type)) do |f| %>
+    <%= render "form_directory", f: f if @product.new_record? %>
```
ericfranz: Will we ever go to new when `@product.new_record?` is not true? Could we just have `<%= render "form_directory" %>`?

config/routes.rb
```diff
@@ -28,6 +28,10 @@
         end
       end
       get 'create_key' => 'products#create_key'
+      get 'new_from_git_remote' => 'products#new_from_git_remote'
+      get 'new_from_rails_template' => 'products#new_from_rails_template'
+      post 'create_from_git_remote' => 'products#create_from_git_remote'
+      post 'create_from_rails_template' => 'products#create_from_rails_template'
```
ericfranz: I think you could do this inside the `resources :products` block:

```ruby
collection do
  get 'new_from_git_remote'
  get 'new_from_rails_template'
  post 'create_from_git_remote'
  post 'create_from_rails_template'
end
```

Then you would end up with the same result.

app/views/products/new_from_rails_template.html.erb
```diff
@@ -0,0 +1,13 @@
+<%= render 'header' %>
+
+<%= bootstrap_form_for(@product, as: :product, url: create_from_rails_template_path(type: @type)) do |f| %>
+    <%= render "form_directory", f: f if @product.new_record? %>
```
ericfranz: Will we ever go to new when `@product.new_record?` is not true? Could we just have `<%= render "form_directory" %>`?

app/models/product.rb
```diff
@@ -14,7 +14,7 @@ class Product
   validates :name, presence: true
 
   validate :app_does_not_exist, on: :create_app
```
ericfranz: This validation is now never run. We should probably change it to:

```ruby
validate :app_does_not_exist, on: [:create_from_git_remote, :create_from_rails_template]
```

app/views/products/new_from_git_remote.html.erb
brianmcmichael: I'm not sure, this was ported from the existing form logic. I'll have to look into it, but at first glance it doesn't appear that it's needed in this view.

config/routes.rb
brianmcmichael: Doing this changes the routes. I've updated the references, but I'm pointing it out because I think it's kind of ugly.

```
                create_key_products GET    /admin/:type/products/create_key(.:format)                               products#create_key {:type=>/dev|usr/}
       new_from_git_remote_products GET    /admin/:type/products/new_from_git_remote(.:format)                      products#new_from_git_remote {:type=>/dev|usr/}
   new_from_rails_template_products GET    /admin/:type/products/new_from_rails_template(.:format)                  products#new_from_rails_template {:type=>/dev|usr/}
    create_from_git_remote_products POST   /admin/:type/products/create_from_git_remote(.:format)                   products#create_from_git_remote {:type=>/dev|usr/}
create_from_rails_template_products POST   /admin/:type/products/create_from_rails_template(.:format)               products#create_from_rails_template {:type=>/dev|usr/}
```

ericfranz: "Doing this changes the routes"

What do the routes look like without this change? i.e. can you show the relavent output of `rake routes` for both?

brianmcmichael: Before:
```         
                create_key GET    /admin/:type/products/create_key(.:format)                               products#create_key {:type=>/dev|usr/}
       new_from_git_remote GET    /admin/:type/products/new_from_git_remote(.:format)                      products#new_from_git {:type=>/dev|usr/}
   new_from_rails_template GET    /admin/:type/products/new_from_rails_template(.:format)                  products#new_from_rails_template {:type=>/dev|usr/}
    create_from_git_remote POST   /admin/:type/products/create_from_git_remote(.:format)                   products#create_from_git_remote {:type=>/dev|usr/}
create_from_rails_template POST   /admin/:type/products/create_from_rails_template(.:format)               products#create_from_rails_template {:type=>/dev|usr/}
```

Note that doing this as a `collection do` block appends `_products` to each of the routes.

ericfranz: Ah you mean the route helpers. The routes are identical.

Solution for route helpers is to use `as:` http://guides.rubyonrails.org/routing.html#naming-routes

So to get what you had before you would have to do:

```
collection do
  get 'new_from_git_remote', as: 'new_from_git_remote'
  # ... etc
```

Sorry.

brianmcmichael: That didn't have any effect. It was still appending `_products` to the named route.

brianmcmichael:
@ericfranz Please review latest batch of changes made at https://github.com/OSC/ood-dashboard/compare/fcc540ce9068...cf1568d9e782

Issue #46, opened 2016-10-17
For Shared Apps, add a new button called "Rebuild App" that does Bundle Install, Rebuild Assets, and Restart App


ericfranz:
_I copied this description from the conversation below, then updated it accordningly._

https://github.com/OSC/ood-dashboard/blob/3080b871f93c7413d528beff02587652fb2d4ec5/app/controllers/products_controller.rb#L110-L115

Add a new command:

```
rebuild_passenger_rails_app: [{"HOME" => "", "RAILS_ENV" => "production"}, "bin/bundle install --path=vendor/bundle && bin/rake assets:clobber && bin/rake assets:precompile && bin/rake tmp:clear && mkdir -p tmp && touch tmp/restart.txt && echo 'Done!'"]
```

And a button after https://github.com/OSC/ood-dashboard/blob/3080b871f93c7413d528beff02587652fb2d4ec5/app/views/products/show.html.erb#L47

`Rebuild App` should be the title of the button and make it the bootstrap "primary" color? Bonus: tooltip should indicate it does all three of the other buttons actions in a row: Bundle Install, Rebuild Assets, and Restart App

Show only `if @type == :usr && @product.passenger_rails_app?`

See below for conversation regarding how we got here.

Original issue description was something like:

>For Shared Apps add Update App button that does that does Bundle Install, Rebuild Assets, and Restart App

nickjer:
Is this called "Update App" for dev apps as well?

ericfranz:
I'd just do it for shared apps, because shared apps its always: fetch/checkout tag/update

But that isn't the case with dev apps

ericfranz:
https://github.com/OSC/ood-dashboard/blob/3080b871f93c7413d528beff02587652fb2d4ec5/app/controllers/products_controller.rb#L110-L115

Add a new command:

```
update_app: [{"HOME" => "", "RAILS_ENV" => "production"}, "bin/bundle install --path=vendor/bundle && bin/rake assets:clobber && bin/rake assets:precompile && bin/rake tmp:clear && mkdir -p tmp && touch tmp/restart.txt && echo 'Done!'"]
```

And a button after https://github.com/OSC/ood-dashboard/blob/3080b871f93c7413d528beff02587652fb2d4ec5/app/views/products/show.html.erb#L47

[...]

Issue #102, opened 2016-12-16
Refactor dashboard code dealing with shared apps and CLI Commands


ericfranz:
This is the current implementation:

1. https://github.com/OSC/ood-dashboard/blob/7c8981a0fc083d17230c595ac9a641bc12778324/app/views/products/show.html.erb#L44-L48
2. https://github.com/OSC/ood-dashboard/blob/7c8981a0fc083d17230c595ac9a641bc12778324/app/controllers/products_controller.rb#L110-L139

Each app type has a different list of possible CLI commands. Instead of having all the checks to determine which commands to build in the view, we could do this in the model if the model had a list of CliCommand objects. The class could have an execute method that accepted the request.stream object as input and perhaps the directory to cd to when executing the command.

This might make the code clearer.

ericfranz:
Not focusing on developer dashboard right now.

PR #111, opened 2017-01-04
Add rebuild app

brianmcmichael:
Implements https://github.com/OSC/ood-dashboard/issues/46

I use the button_tag here directly because the helper isn't flexible enough. If you want I can go ahead and modify the `command_btn()` helper to cut a few chars off this line.


ericfranz:
> I use the button_tag here directly because the helper isn't flexible enough

It wasn't apparent just from eyeing the method, but how does the command_btn helper need to change to support this? The color of the command_btn? Go ahead.

Also, the display command using scl with multiple commands needs to have the `-- <<<` i.e.:

```diff
-scl enable rh-ruby22 nodejs010 --- 'command1 && command2 && command3'
+scl enable rh-ruby22 nodejs010 -- <<< 'command1 && command2 && command3'
```

ericfranz:
I guess the command_btn helper should wrap `<code></code>` tags around the display argument so that these don't need to be added to each value. Or you can keep it but make sure to add the missing tags to the precompile assets, which I missed.

One of the ideas of this helper was separating the view code from the domain specific code, to make #102 easier to implement. You could imagine "title", "key", and "display" being attributes on a CliCommand model object, and the view code simply iterating over this command object. In that case, maybe command_btn would accept an argument like `primary: false` so if you set `primary: true`. Or something like that. Just consider whatever you add that, this value/property in the future might be initially set on a model object by a Product model instance (i.e. such as lazily built when calling a getter `@product.cli_commands`)

brianmcmichael:
I'd want to update the parameters in `command_btn` to more closely track `button_tag`, for example, one of the keys used on `command_btn` is `title`, but that gets translated to an html tag called `data-title`. This is specifically problematic because it conflicts with the `title` html tag that provides the hover-over tooltip.

It might also be better to allow any input hash options on `command_btn`, then convert additional hash keys to html element tags like `button_tag`

http://apidock.com/rails/ActionView/Helpers/FormTagHelper/button_tag
```
Options

:data - This option can be used to add custom data attributes.

:disabled - If true, the user will not be able to use this input.

Any other key creates standard HTML options for the tag.
```

brianmcmichael:
I'm not sure what purpose the `-- <<<` serves. I notice that it also isn't on the "Bundle Install" button. Should it be added there?

This seems like another situation where having the actual command available would be handy.

brianmcmichael:
Nevermind `-- <<<` is for multiple commands. Got it.

ericfranz:
`command_btn` was meant to take a future class `CliCommand` attributes as arguments and produce a button for it; so these translations make sense:

* `CliCommand#title` => html data-title
* `CliCommand#help` => html title (for tooltip?)

>It might also be better to allow any input hash options on command_btn, then convert additional hash keys to html element tags like button_tag

Maybe but if we start varying the way we display these buttons too much from each other then we lose the value of having a `command_btn` helper. I didn't want to get crazy with these buttons, only add functionality as needed to the helper.

ericfranz:
So to clarify, eventually the goal was that none of the command buttons would be hardcoded in the view.

brianmcmichael:
Button and helper have been updated. @ericfranz please review.

app/helpers/products_helper.rb
```diff
@@ -34,14 +34,22 @@ def ssh_key
...
   end
 
-  def command_btn(title:, key:, display:)
+  # A custom button helper for command line tools
+  #
+  # @param [String] title: The button label
+  # @param [String] key: The key of the command in the CMDS hash ( See: controllers/products_controller.rb )
+  # @param [String] display: The equivalent command to be displayed to the user
+  # @param [optional, String] help: Tooltip text to be provided on hover. ( Default: none )
+  # @param [optional, String] color: The bootstrap color of the button. Ex. "primary", "info", "success", etc. ( Default: "default" )
```
nickjer: These don't follow Yardoc specifications for `@param` (see http://www.rubydoc.info/gems/yard/file/docs/Tags.md#param)

brianmcmichael: Thanks, I was using https://gist.github.com/chetan/1827484 for formatting guidance, which lists the name after the type. I'll go with the official docs.

nickjer:
Copying and pasting:

```
RAILS_ENV=production scl enable rh-ruby22 nodejs010 -- <<< 'bin/bundle install --path=vendor/bundle && bin/rake assets:clobber && bin/rake assets:precompile && bin/rake tmp:clear && mkdir -p tmp && touch tmp/restart.txt && echo 'Done!''
```

Does not work. I get:

```
bash: !'': event not found
```

nickjer:
Tooltip says "Rebuild Assets" but the button says "Precompile Assets".

brianmcmichael:
> `bash: !'': event not found`

This was fixed by escaping the `!` in the echo statement. 

> Tooltip says "Rebuild Assets" but the button says "Precompile Assets".

Where are you seeing this? There shouldn't be a tooltip on the "Precompile Assets" button.

nickjer:
> Where are you seeing this?

Forgot to mention I am seeing it on the Rebuild App button.

https://github.com/OSC/ood-dashboard/blob/05d22fbecc1a6406cc1daa498342cd4780d158d5/app/views/products/show.html.erb#L48

brianmcmichael:
This is what I've got. What browser are you using?

rebuildtooltip

nickjer:
> Tooltip says "Rebuild Assets" but the button says "Precompile Assets".

In your image it says "Rebuild Assets", but the button says "Precompile Assets". So the tooltip needs to be updated with the correct text.

brianmcmichael:
OK, I've updated that, but I don't know that it's an ideal statement because it doesn't express the fact that we're performing an `assets:clobber` prior to the precompilation. That may be a problem with the Precompile button as well.

nickjer:
If we want to rename the button we should open that up in a separate issue. As the name "Precompile Assets" exists in multiple places through out the code.

PR #114, opened 2017-01-13
add dashboard site-wide announcement

ericfranz:
Addresses https://github.com/OSC/ood_appkit/issues/32 but just for the dashboard.

Create a file at `/etc/ood/config/announcement.md` and the contents will be displayed as a site-wide message to everybody.

This is a candidate for a feature that could be useful on a per-app basis as well.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -3,7 +3,7 @@ class ApplicationController < ActionController::Base
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
 
-  before_filter :set_user, :set_logout_url, :set_nav_groups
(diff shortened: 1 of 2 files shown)


nickjer:
The reason I would lean towards a yaml or json format is so we can attach more context to the messages. For example we could have something like:

```yaml
- begins: 6464894651
  expires: 6464898000
  alert: 'notice'
  icon: fa://exclamation
  msg: 'Zombies have invaded! Please take shelter immediately.`
- ...
```

brianmcmichael:
I'd lean json for external files just because it's a little better supported in other (non-Ruby) languages.

ericfranz:
We can change it to yaml or json later if we need more context.

PR #134, opened 2017-03-03
add spacing below row. fixes #132

### LINKED ISSUE #132: Spacing below "Rebuild App" button
When there are app errors, the "Rebuild App" button abuts the error panel. It just needs a few px of spacing here.

spacing

brianmcmichael:
I borrowed the new spacing classes from bootstrap 4 for this:
https://v4-alpha.getbootstrap.com/utilities/spacing/

If/when this is ever updated to Bootstrap 4 the custom style can be removed and it will work the same way.

### DIFF
--- a/app/assets/stylesheets/application.css.scss
+++ b/app/assets/stylesheets/application.css.scss
@@ -51,6 +51,12 @@ body {
   padding-top: $navbar-height + 10;
 }
 
+// Borrowed from bootstrap 4
+// Remove this block if/when upgraded to bootstrap 4
+.mb-2 {
+  margin-bottom: (2rem * .25) !important;
+}
+
 // Your custom css styles below (must be manually added)
 @import "dashboard";
 @import "navbar";
--- a/app/views/products/show.html.erb
+++ b/app/views/products/show.html.erb
@@ -9,7 +9,7 @@
 
 <h3>Details</h3>
 
-<div class="row">
+<div class="row mb-2">
   <div class="col-md-2 col-sm-2 col-xs-12">
     <p>
       <%= app_icon_tag(@product.app) %>


nickjer:
What does it look like now with & without the warning box below.

brianmcmichael:
With warning now:

after

Without warning now:

nodiv

nickjer:
As an aside, these are the browsers that support `rem`:

http://caniuse.com/#feat=rem

app/assets/stylesheets/application.css.scss
```diff
@@ -51,6 +51,12 @@ body {
   padding-top: $navbar-height + 10;
 }
 
+// Borrowed from bootstrap 4
+// Remove this block if/when upgraded to bootstrap 4
+.mb-2 {
+  margin-bottom: (2rem * .25) !important;
```
nickjer: Does the 0.25 correspond to 2 or 1 in `.mb-2`?

nickjer:
I think we should rename this to something that doesn't collide with bootstrap v4, as if/when we do update we won't see odd behavior.

brianmcmichael:
The math was off, but it should be the same product. Should be `(1rem * .50)` `for mb-2`
https://github.com/twbs/bootstrap/blob/d79b5563730e8188ca7ee5dfc71503d82eb297b4/scss/_variables.scss#L141-L169

nickjer:
:+1:

ericfranz:
This is the wrong solution.

We have

```erb
<h3>Details</h3>
<div class="row">
  <%# show details with build buttons %>
</div>

<% unless @product.valid?(:show_app) %>
<div class="alert alert-danger">
  <%# show error content %>
</div>
<% end %>


<% if @product.permissions? %>
<h3>Permissions</h3>
<div class="row">
    <%# show details with build buttons %>
</div>
<% end %>
```

We should wrap the alert in another row.

```diff
<h3>Details</h3>
<div class="row">
  <%# show details with build buttons %>
</div>

<% unless @product.valid?(:show_app) %>
+<div class="row">
+<div class="col-md-12">
<div class="alert alert-danger">
  <%# show error content %>
</div>
+</div>
+</div>
<% end %>


<% if @product.permissions? %>
<h3>Permissions</h3>
<div class="row">
    <%# show details with build buttons %>
</div>
<% end %>
```

brianmcmichael:
@ericfranz I disagree that this is the wrong solution. That was my first thought but it doesn't work here. Neither the bootstrap `"row"` or `"col-md-x"` provide top or bottom padding.

bootstrap

Bootstrap relies on element spacing to keep things separated. The `alert` element provides 20px of bottom margin, but none on the top. Buttons have top margin, but none on the bottom. In bootstrap 3 the solution seems to be that we need to add custom spacing to keep these apart, which I have done here by borrowing from bootstrap 4.

PR #147, opened 2017-04-10
Update setup script

nickjer:
Uses latest Rails `bin/setup` for inspiration.

- drops out when error encountered on command line
- accepts `PASSENGER_APP_ENV` and `PASSENGER_BASE_URI`
- compiles assets if run in production mode

It will build the app from a first-time install, and can be used to update the app if checking out a new version.


bin/setup
```diff
@@ -1,29 +1,40 @@
 #!/usr/bin/env ruby
-require 'pathname'
+require "pathname"
+require "fileutils"
+
+include FileUtils
+
+ENV["RAILS_ENV"]               ||= ENV["PASSENGER_APP_ENV"]
+ENV["RAILS_RELATIVE_URL_ROOT"] ||= ENV["PASSENGER_BASE_URI"]
```
ericfranz: This sets ENV["RAILS_ENV"] to ENV["PASSENGER_APP_ENV"] no matter what. So if PASSENGER_APP_ENV is not set, `RAILS_ENV` will exist and be set to `""`. Seems more appropriate that if RAILS_ENV or PASSENGER_APP_ENV are not set, we don't create the environment at all.

```ruby
ENV["RAILS_ENV"]               ||= ENV["PASSENGER_APP_ENV"] if ENV["PASSENGER_APP_ENV"]
ENV["RAILS_RELATIVE_URL_ROOT"] ||= ENV["PASSENGER_BASE_URI"] if ENV["PASSENGER_BASE_URI"]
```

bin/setup
```diff
@@ -1,29 +1,40 @@
...
+  puts "== Building Dashboard App =="
+  puts "RAILS_ENV               = #{ENV["RAILS_ENV"]}"
+  puts "RAILS_RELATIVE_URL_ROOT = #{ENV["RAILS_RELATIVE_URL_ROOT"]}"
 
-  # puts "\n== Copying sample files =="
-  # unless File.exist?("config/database.yml")
-  #   system "cp config/database.yml.sample config/database.yml"
-  # end
+  puts "\n== Installing dependencies =="
+  system("bin/bundle check &> /dev/null") || system!("bin/bundle install --deployment")
```
ericfranz: Would we run `bin/bundle install --deployment` in development? I know that this does the equivalent of `bundle install --path=vendor/bundle` but it does a few other things, but I haven't looked at it enough to understand.

bin/setup
ericfranz: Actually, we should use a sensible default for RAILS_ENV. If PASSENGER_APP_ENV is not set, we need to set the default. Either default to production or development. Rails convention is development, but I think we should default to production. Or forget defaults, and exit with error if `PASSENGER_APP_ENV` or `RAILS_ENV` is not specified. But don't throw an error about `RAILS_RELATIVE_URL_ROOT` since that is normally specified in the dotenv file.

bin/setup
```diff
@@ -1,29 +1,40 @@
...
+chdir APP_ROOT do
   # This script is a starting point to setup your application.
   # Add necessary setup steps to this file:
 
-  puts "== Installing dependencies =="
-  system "gem install bundler --conservative"
-  system "bundle check || bundle install"
+  puts "== Building Dashboard App =="
+  puts "RAILS_ENV               = #{ENV["RAILS_ENV"]}"
+  puts "RAILS_RELATIVE_URL_ROOT = #{ENV["RAILS_RELATIVE_URL_ROOT"]}"
```
ericfranz: Display a different message if `RAILS_RELATIVE_URL_ROOT` not set? Something like:

>RAILS_RELATIVE_URL_ROOT not specified; rake tasks will use whatever is defined in dotenv files (if using dotenv) or framework default

Or just omit if not set?

bin/setup
nickjer: Setting an environment variable to `nil` in Ruby unsets the environment variable. So if both `RAILS_ENV` and `PASSENGER_APP_ENV` are unset, then `RAILS_ENV` stays unset.

bin/setup
nickjer: I am fine with both, I guess I can not show it if it is unset for now.

bin/setup
nickjer: From the site:

> The `--deployment` flag activates a number of deployment-friendly conventions:
>
> Isolate all gems into `vendor/bundle`
> Require an up-to-date `Gemfile.lock`
> If `bundle package` was run, do not fetch gems from rubygems.org. Instead, only use gems in the checked in `vendor/cache`

I think we want the first two. Since these are Rails apps we don't need or support `bundle package`.

nickjer:
I forgot about the `OOD_SITE` and `OOD_PORTAL`, @brianmcmichael reminded me just before I saw this.

nickjer:
@ericfranz let me know if you are happy with these changes, then I will do something similar for the following apps (taking into consideration the NodeJS apps):

- shell
- files
- file editor
- my jobs
- active jobs

And update the ood-apps-installer script to take advantage of this and be called as:

```sh
# Default OOD install
scl enable rh-ruby22 nodejs010 git19 -- rake

# OSC OnDemand
scl enable rh-ruby22 nodejs010 git19 -- rake OOD_SITE=osc OOD_PORTAL=ondemand

# OSC AweSim
scl enable rh-ruby22 nodejs010 git19 -- rake OOD_SITE=osc OOD_PORTAL=awesim
```

ericfranz:
To be clear, the last three commands you gave examples of were running the old installer, not running only this bin/setup script

nickjer:
The last three commands I gave were examples of the new ood-apps-installer code. It would download and build all the defined apps.

nickjer:
Also, we don't expect anyone besides maybe a developer to run `bin/setup`. Every sys admin should be running the `ood-apps-installer` utility for installing/updating the apps.

nickjer:
I think the default environment should be `RAILS_ENV=development`.

As all production cases for Open OnDemand will use the app installer script, which will always define `PASSENGER_APP_ENV=production`. Or if they deploy a shared app from the dashboard, and that is an automated process where we also set `RAILS_ENV=production`.

So the only case where I think a user might manually call `bin/setup` is when they are setting up a development version of the app.

ericfranz:
I'm okay with defaulting to development.

Issue #150, opened 2017-04-27
Display MOTD even if unformatted


nickjer:
Would it be possible to display `/etc/motd` if it is unformatted. Similar to the way we did it in the past with just a `<pre></pre>` tag? This would then make the MOTD option available to most centers right away.

ericfranz:
This feature needs a proposed design. Just attach it to this issue.

1. What environment variable for configuration will be used? MOTD_FORMAT probably makes sense to go along with MOTD_PATH?
2. Do we want to use the default <pre>. Bootstrap places that in a well, and it might look ugly to have a giant well in the center of the front page. We might want to customize styles in this case.
3. Since we will have multiple MOTD formatters, we will want to be able to use different formatters based on the specified format. This will let us easily add an RSS format in the future. So how will the code change to accommodate this?

ericfranz:
For reference: https://github.com/OSC/ood-dashboard/wiki/Message-of-the-Day

ericfranz:
Suggested formats:

* osc
* markdown
* plaintext
* rss

brianmcmichael:
#151 has been merged

PR #151, opened 2017-05-05
Motd options

brianmcmichael:
Proposed initial solution to satisfy https://github.com/OSC/ood-dashboard/issues/150

#### Note: This PR does not address RSS

For context, a default installation does not include any MOTD display by default:

no_motd_path

*****

To display anything, the admin must first supply `ENV['MOTD_PATH']`

The app now defaults to plain text formatting:

motd_path_no_motd_format

*****

To parse the content of the MOTD, the admin can provide `ENV['MOTD_FORMAT']`

There are two options currently, `MOTD_FORMAT='markdown'` and `MOTD_FORMAT='osc'`

### Markdown Style

motd_path_motd_format_markdown

*****

### OSC Style

motd_path_motd_format_osc

*****

### DIFF
--- a/app/assets/stylesheets/application.css.scss
+++ b/app/assets/stylesheets/application.css.scss
@@ -56,8 +56,15 @@ body {
   margin-bottom: (1rem * .50) !important;
 }
 
+pre.motd_monospaced {
+  background-color: #FFFFFF;
+  border: none;
+}
+
 // Your custom css styles below (must be manually added)
 @import "dashboard";
 @import "navbar";
 @import "apps";
 @import "products";
+
+
--- a/app/assets/stylesheets/application.css.scss
+++ b/app/assets/stylesheets/application.css.scss
@@ -57,7 +57,7 @@ body {
 }
 
 // Remove the well-styling from the bootstrap <pre>
-pre.motd_monospaced {
+pre.motd-monospaced {
   background-color: #FFFFFF;
   border: none;
 }
--- a/app/assets/stylesheets/application.css.scss
+++ b/app/assets/stylesheets/application.css.scss
@@ -59,6 +59,7 @@ body {
 // Remove the well-styling from the bootstrap <pre>
 pre.motd-monospaced {
   background-color: #FFFFFF;
+  font-size: 14px;
   border: none;
 }
 
--- a/app/assets/stylesheets/application.css.scss
+++ b/app/assets/stylesheets/application.css.scss
@@ -56,6 +56,7 @@ body {
   margin-bottom: (1rem * .50) !important;
 }
 
+// Remove the well-styling from the bootstrap <pre>
 pre.motd_monospaced {
   background-color: #FFFFFF;
   border: none;
--- a/app/models/motd_file.rb
+++ b/app/models/motd_file.rb
@@ -24,14 +24,6 @@ class MotdFile
     @motd_text_format = format
   end
 
-  # An empty file whose modification timestamp indicates the last time the user
-  # viewed the motd messages. This is useful for when we want to use the file
-  # system to determine when new messages the user has not seen have been
-  # added to the motd.
-  def motd_config_file
-    @motd_config_file ||= OodAppkit.dataroot.join(".motd")
-  end
-
   def exist?
     motd_system_file && File.file?(motd_system_file)
   end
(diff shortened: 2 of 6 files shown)


ericfranz:
The pre text for plain text formatting for the motd should have the font size set to 14px as it is at 13 px which makes it smaller than the others.

Not sure, should we add the "Message of the Day" title to these sections regardless of the formatting? I notice we only show it if its OSC format. That said, I remember Dave mentioning it odd we called ours "Message of the Day" since most of the messages are more than a day old, some times weeks or months ha ha.

brianmcmichael:
I deliberately removed the big "Message of the day" header from the Plaintext and Markdown versions because the test markdown file I was using actually began with `## Message of the Day` and it ended up showing up twice when I still had it there. I figured we probably don't want to proscribe content for other sites.

ericfranz:
Make three classes:

```
class MotdFormatterMarkdown
  def to_partial_path
    "dashboard/motd_markdown"
  end
end

class MotdFormatterPlainText
  def to_partial_path
    "dashboard/motd_text"
  end
end

class MotdFormatterOsc
  def to_partial_path
    "dashboard/motd_osc"
  end

  # since this actually splits the file content into separate messages
  # we can still use this as it is
  def messages
    # original implementation
  end
end
```

If in the dashboard controller https://github.com/OSC/ood-dashboard/blob/f829805e0bed96c4b3bcb88bc514bf44614fdf14/app/controllers/dashboard_controller.rb#L6 `@motd_file` is actually one of these formatter instances, you can then do this in the view (https://github.com/OSC/ood-dashboard/blob/f829805e0bed96c4b3bcb88bc514bf44614fdf14/app/views/dashboard/index.html.erb#L6):

```diff
-<%= render partial: "motd", locals: { motd: @motd_file  } if @motd_file.exist? %>
+<%= render @motd_file if @motd_file.exist? %>
```

Though I think that if you do this, in the partial the local object will be set to the name of the partial. i.e. in the motd_osc partial you would use `motd_osc` object and call `motd_osc.messages` on it, whereas in the `motd_text` partial the object would be `motd_text`. I think.

http://api.rubyonrails.org/classes/ActionView/PartialRenderer.html

ericfranz:
That comment is pseudo code of course, you would have to have access to the file contents and handle the case that the file ! exist, as before.

nickjer:
Note that `ood_appkit` comes with a custom configured markdown handler:

https://github.com/OSC/ood_appkit#markdown-handler

So you can use `OodAppkit.markdown`.

Also to take advantage of the styles provided in `ood_appkit` that mimic GitHub styles, you want to enclose the rendered text with the classes `class="ood-appkit markdown"` (similar to what we do here in the [show.html.erb](https://github.com/OSC/ood_appkit/blob/1c8bae0f21f663af0ba87ae192315dd37e193426/app/views/ood_appkit/wiki/show.html.erb)).

nickjer:
Also, I'd maintain the formatting for HTML classes's used in the app already and use hyphen's as opposed to snakecase.

For example, changing: `motd_monospaced` to `motd-monospaced` since it is an HTML class.

brianmcmichael:
I understand that there's a parser in `ood_appkit`, but as a developer I don't understand why I would choose it over the built-in Redcarpet functionality, especially because I have complete control over the latter.

nickjer:
OodAppkit uses RedCarpet for the markdown renderer as well. It is also specifically curated to mimic GitHub markdown as best as possible. You can also easily override it in an initializer.

If you use it you will also have a consistent markdown renderer across the entire dashboard. As opposed to having some markdown support in the motd, but a different feature set elsewhere in the code.

ericfranz:
FWIW I just created this issue: https://github.com/OSC/ood_appkit/issues/43
So in the future probably we will replace redcarpet with kramdown.

brianmcmichael:
Is there a reason for switching?
https://ruby.libhunt.com/project/kramdown/vs/redcarpet

brianmcmichael:
Temporarily assigned to @ericfranz for example case of model refactor per offline discussion.

brianmcmichael:
Changes described at https://github.com/OSC/ood-dashboard/pull/151#issuecomment-299612645 are available at https://github.com/OSC/ood-dashboard/pull/152

I'm sure that the factory method in the controller can be cleaned up and moved out, but I'm not sure where else to put it.

ericfranz:
Doing https://github.com/OSC/ood-dashboard/pull/152 instead

PR #152, opened 2017-05-12
Utilize to_partial_path in rendering motd

brianmcmichael:
Alternative PR branched from https://github.com/OSC/ood-dashboard/pull/151

Uses `to_partial_path` as described at https://github.com/OSC/ood-dashboard/pull/151#issuecomment-299612645

This is a pretty heavy haircut of #151 so I put is as a separate branch for easier comparison. Close #151 if this is merged instead.


app/views/dashboard/index.html.erb
```diff
@@ -3,7 +3,7 @@
 <p class="lead">OnDemand provides an integrated, single access point for all of your HPC resources.</p>
 <br>
 
-<%= render partial: "motd", locals: { motd: @motd_file  } if @motd_file.exist? %>
+<%= render @motd_file if @motd_file.exist? %>
```
ericfranz: Add comment saying "file specifies partial path" for rendering purpose

app/views/dashboard/_motd.html.erb
```diff
@@ -1,18 +1 @@
...
-
-<% end.empty? %>
-
-  <hr>
-  <div class="motd">
-    <h4 class="motd_title">There are no new messages.</h4>
-  </div>
-
-<% end %>
+<%= render motd_file %>
```
ericfranz: Can this file be deleted? Is it a duplicate of another? Is this used anywhere else?

app/controllers/dashboard_controller.rb
```diff
@@ -3,7 +3,14 @@ def index
...
 
-    @motd_file = MotdFile.new
+    @motd_file = case ENV['MOTD_FORMAT']
+    when 'osc'
+      MotdFile::MotdFormatterOsc.new
+    when 'markdown'
+      MotdFile::MotdFormatterMarkdown.new
+    else
+      MotdFile::MotdFormatterPlainText.new
+    end
```
ericfranz: Instead of inheritance, I'd consider:

```
file = MotdFile.new(ENV["MOTD_PATH")
formatted_file = MotdFile::MotdFormatterOsc.new(file)
```

i.e. instead of 

```ruby
class MotdFormatterPlainText < MotdFile
  def initialize
    super
    @content = motd_content
  end

  def to_partial_path
    "dashboard/motd_text"
  end
end
```

something like

```ruby
class MotdFormatterPlainText
  def initialize(motd_file)
    @content = motd_file.content
  end

  def to_partial_path
    "dashboard/motd_text"
  end
end
```

ericfranz: Maybe there is a URI class that has a content method on it?

ericfranz:
Also update the osc and awesim specfic .env files to specify the MOTD_FORMAT etc.

app/models/motd_file.rb
```diff
@@ -1,67 +1,43 @@
...
+  #
+  # Uses open-uri to check local or remote path for contents
   def exist?
-    motd_system_file && File.file?(motd_system_file)
+    exists = false
+    if motd_uri
+      begin
+        open(motd_uri)
+        exists = true
+      rescue
```
nickjer: Not sure we should rescue from everything and assume it is because the file/url doesn't exist. Could be something like `motd_uri` is an undefined method.

app/models/motd_file.rb
```diff
@@ -1,67 +1,43 @@
...
+        # The messages file does not exist.
+        Rails.logger.warn "MOTD File is missing; it was expected at #{motd_uri}"
```
nickjer: We should also output the error message to have a better understanding of what broke. For example it may be due to an invalid SSL certificate.

app/models/motd_file.rb
```diff
@@ -1,67 +1,43 @@
...
+    exists
   end
 
-  # If the motd file hasn't been created on the system, or if the system motd is newer than the user's file, return true.
-  def new_messages?
-    # FIXME: Use if/else statements because this is arcane
-    (messages.count > 0) ? ( !File.exist?(motd_config_file) ? true : File.new(motd_system_file).ctime > File.new(motd_config_file).ctime ? true : false ) : false   
+  # Create an array of message objects based on the current message of the day.
+  def content
+    if self.exist?
```
nickjer: This is the only place that `#exist?` is used. You can get rid of `exist?` and simplify this to:

```ruby
def content
  motd_uri ? open(motd_uri).read : nil
rescue ERROR_HERE => e
  Rails.logger.warn "..."
  nil
end
```

You'd then check if `content` exists or not in the formatters. Don't forget to convert to a string or fallback to defaults in the formatters if content is `nil`.

app/models/motd_formatter_osc.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  def to_partial_path
+    "dashboard/motd_osc"
+  end
+
+  # since this actually splits the file content into separate messages
+  # we can still use this as it is
+  def messages
+    if content
```
nickjer: According to your code this will always be true.

app/views/dashboard/index.html.erb
```diff
@@ -3,7 +3,8 @@
 <p class="lead">OnDemand provides an integrated, single access point for all of your HPC resources.</p>
 <br>
 
-<%= render partial: "motd", locals: { motd: @motd_file  } if @motd_file.exist? %>
+<%# This assumes @motd_file responds to `to_partial_path` %>
+<%= render @motd if @motd %>
```
nickjer: This will always be true. In fact if they do not set the env var `MOTD_PATH`, it will render:

```html
<pre class="motd-monospaced"></pre>
```

This isn't a problem, but was just wondering if this is what you intended.

app/models/motd_file.rb
brianmcmichael: `open-uri` can return an IOError, a HTTPError, or a FTPError, depending on what you're trying to open. 

If `motd_uri` is not defined then that's a bug in our code. Maybe this should just be fixed by making the error message a little more clear.

app/models/motd_file.rb
brianmcmichael: I don't know how I feel about this logging at all, actually. I worry that if a site doesn't set up a `motd` we're going to be bombarding their logs with error messages about motd files not present every time a user hits the dashboard.

It may be better to catch and move on from expected IO, HTTP, and FTP errors and only log errors that don't match, or when `motd_uri` isn't blank.

app/models/motd_file.rb
nickjer: It is still safer to do:

```ruby
rescue IOError, HTTPError, FTPError => e
```

app/models/motd_file.rb
brianmcmichael: https://github.com/OSC/ood-dashboard/blob/df0d4a1c6b77b9615c558adf4f491c6a1b87e4db/app/views/dashboard/index.html.erb#L6

and

https://github.com/OSC/ood-dashboard/blob/ce5b35217bb626aeffbc79dd97bc1be177220b07/app/views/apps/index.html.erb#L14

app/models/motd_file.rb
nickjer: But yeah, if you want to rescue all `StandardError`'s, then at least get the error message logged as you mentioned.

app/models/motd_file.rb
nickjer: The first link doesn't exist in this PR anymore:

https://github.com/OSC/ood-dashboard/blob/e68a8e334b60cafd817ec118a6a5cb9336d0fd7d/app/views/dashboard/index.html.erb#L7

and the second link will actually be a bug for this PR.

app/models/motd_file.rb
```diff
@@ -1,67 +1,36 @@
...
+  def load(motd_uri)
+    motd_uri ? open(motd_uri).read : nil
+  rescue Errno::ENOENT
+    Rails.logger.warn "MOTD File is missing; it was expected at #{motd_uri}"
+    nil
+  rescue OpenURI::HTTPError
+    Rails.logger.warn "MOTD File is not available at #{motd_uri}"
+    nil
+  rescue StandardError => ex
+    Rails.logger.warn "Error opening MOTD at #{motd_uri}\nException: #{ex}"
```
nickjer: I think this should be `ex.message`.

nickjer:
I think you need to go through and confirm that all the Formatters can handle `nil` for `motd.content`.

app/models/motd_formatter_osc.rb
ericfranz: motd.content can return nil; then this would be false. `#messages` should probably return an empty array if the content is nil, however

ericfranz:
We could just ensure that MotdFile#content always returns "" if @content is set to nil:

```
def content
  content || ""
end
```

Then the other classes that use this can be confident that content is nil.

app/models/motd_formatter_rss.rb
```diff
@@ -0,0 +1,16 @@
+require 'rss'
+
+class MotdFormatterRss
+
+  attr_reader :content
+
+  # @param [MotdFile] motd_file an MotdFile object that contains a URI path to a message of the day in OSC format
+  def initialize(motd_file)
+    feed = motd_file.content
+    @content = RSS::Parser.parse(feed)
```
ericfranz: 1. This can throw exceptions.

```
irb(main):009:0> RSS::Parser.parse "<Ddsak<dsalkl>"
RSS::NotWellFormedError: This is not well formed XML
malformed XML: missing tag start
Line: 1
Position: 14
Last 80 unconsumed characters:
<Ddsak<dsalkl>
        from /opt/rh/rh-ruby22/root/usr/share/ruby/rss/rexmlparser.rb:19:in `rescue in _parse'
        from /opt/rh/rh-ruby22/root/usr/share/ruby/rss/rexmlparser.rb:16:in `_parse'
        from /opt/rh/rh-ruby22/root/usr/share/ruby/rss/parser.rb:164:in `parse'
        from /opt/rh/rh-ruby22/root/usr/share/ruby/rss/parser.rb:79:in `parse'
        from (irb):9
        from /users/PZS0562/efranz/ondemand/dev/dashboard/vendor/bundle/ruby/gems/railties-4.2.7.1/lib/rails/commands/console.rb:110:in `start'
        from /users/PZS0562/efranz/ondemand/dev/dashboard/vendor/bundle/ruby/gems/railties-4.2.7.1/lib/rails/commands/console.rb:9:in `start'
        from /users/PZS0562/efranz/ondemand/dev/dashboard/vendor/bundle/ruby/gems/railties-4.2.7.1/lib/rails/commands/commands_tasks.rb:68:in `console'
        from /users/PZS0562/efranz/ondemand/dev/dashboard/vendor/bundle/ruby/gems/railties-4.2.7.1/lib/rails/commands/commands_tasks.rb:39:in `run_command!'
        from /users/PZS0562/efranz/ondemand/dev/dashboard/vendor/bundle/ruby/gems/railties-4.2.7.1/lib/rails/commands.rb:17:in `<top (required)>'
        from bin/rails:4:in `require'
        from bin/rails:4:in `<main>'
```

2. Also this assumes the RSS descriptions are formatted HTML. But in our case the format should be plain text that is markdown compatible:

https://github.com/OSC/rss_to_motd/blob/eed45d5338683dc276a6f02910a18cee6e50e59f/test/fixtures/motd_rss_example.xml

But I guess some sites might have just normal html descriptions... so maybe this is yet another formatter

brianmcmichael: This is in progress.

app/controllers/apps_controller.rb
```diff
@@ -29,7 +29,17 @@ def index
...
+      case ENV['MOTD_FORMAT']
+        when 'osc'
+          @motd = MotdFormatterOsc.new(motd_file)
+        when 'markdown'
+          @motd = MotdFormatterMarkdown.new(motd_file)
+        when 'rss'
+          @motd = MotdFormatterRss.new(motd_file)
+        else
+          @motd = MotdFormatterPlaintext.new(motd_file)
+      end
```
ericfranz: Why do we render the MOTD file here?

brianmcmichael: If `MOTD_PATH` is not set, nothing will be displayed on the dashboard, but you're right, this should probably be prevented from getting to that point.

brianmcmichael: Added `end if motd_file.exist?`, which ensures the `@motd` is not set if the `ENV['MOTD_PATH']` is `nil` or invalid.

brianmcmichael:
Ok, I think the latest batch of suggestions have been implemented. I'm kicking this over to "needs reviewed" again.

PR #155, opened 2017-05-22
Integrate iHPC support into the dashboard

nickjer:
### After deploying to your AweSim Dev folder, the example iHPC forms can be accessed by:

- Desktops
  - [Oakley](https://apps-dev.awesim.org/pun/dev/ood-dashboard/batch_connect/dev/desktop/oakley/session_contexts/new)
  - [Oakley VDI](https://apps-dev.awesim.org/pun/dev/ood-dashboard/batch_connect/dev/desktop/oakley-vdi/session_contexts/new)
  - [Ruby](https://apps-dev.awesim.org/pun/dev/ood-dashboard/batch_connect/dev/desktop/ruby/session_contexts/new)
  - [Ruby VDI](https://apps-dev.awesim.org/pun/dev/ood-dashboard/batch_connect/dev/desktop/ruby-vdi/session_contexts/new)
  - [Owens](https://apps-dev.awesim.org/pun/dev/ood-dashboard/batch_connect/dev/desktop/owens/session_contexts/new)
  - [Owens VDI](https://apps-dev.awesim.org/pun/dev/ood-dashboard/batch_connect/dev/desktop/owens-vdi/session_contexts/new)
- Abaqus/CAE
  - [Abaqus Software](https://apps-dev.awesim.org/pun/dev/ood-dashboard/batch_connect/dev/abaqus/abaqus-software/session_contexts/new)
  - [Abaqus Hardware](https://apps-dev.awesim.org/pun/dev/ood-dashboard/batch_connect/dev/abaqus/abaqus-hardware/session_contexts/new)
- Jupyter Notebook
  - [Jupyter](https://apps-dev.awesim.org/pun/dev/ood-dashboard/batch_connect/dev/jupyter/jupyter/session_contexts/new)
  - [Jupyter + CUDA](https://apps-dev.awesim.org/pun/dev/ood-dashboard/batch_connect/dev/jupyter/jupyter-cuda/session_contexts/new)

### You can view all iHPC sessions from the following url:

https://apps-dev.awesim.org/pun/dev/ood-dashboard/batch_connect/sessions

### Future work:

  - add Dashboard dropdown menus
  - how do we handle user selection of sub-apps? In dropdown or from an app's menu?
  - strip out custom `ood_core` code and put it into the [ood_core](https://github.com/OSC/ood_core/) library
  - add cluster config support for the `ood_core` component
  - remove stubbing that overrides app routers and looks for apps under `./jobs/...`


ericfranz:
Instead of this:

```ruby
    def token
      "#{router.token}#{"/#{sub_app}" if sub_app}"
    end
```

This might be a little more readable:

```ruby
    def token
      [router.token, sub_app].compact.join("/")
    end
```

ericfranz:
Questions about `SmartAttributes::Attribute`:

1. Why does `SmartAttributes::Attribute#opts` exist? Is this for widget specific arguments? 


2. This still seems odd to me that the initializer is deciding whether or not to set the value to "fixed" based on the argument type. So I can write code that does this:

```ruby
Attribute.new(:custom, value: "default value") # => fixed is false
Attribute.new(:custom, "default value") # => fixed is true
```

Whoever is creating an Attribute should make the decision of what fixed should be, not the Attribute object itself, and there should be only one way to specify the default value. Either:

```ruby
Attribute.new(:custom, value: "default value") # fixed => true by default
Attribute.new(:custom, "default value") # fixed => true by default
```

This doesn't mean that changing this should have an impact on what is in the YAML either.

ericfranz:
In other words, I'd move that "decision" to the code that reads the yaml and instantiates the Attribute objects based on what is in the YAMLs.

ericfranz:
It probably makes the most sense that the default is fixed is false.

So to create an attribute that the user cannot change, I'd one of these:

```ruby
Attribute.new(:custom, "default value", fixed: true)
Attribute.new(:custom, value: "default value", fixed: true)
```

nickjer:
> Why does `SmartAttributes::Attribute#opts` exist? Is this for widget specific arguments?

Yes.

nickjer:
In actuality I could probably get away just storing `@opts`, then using that for the getters of the various pieces. For example instead of:

```ruby
    def widget
      (@widget || "text_field").to_s
    end
```

I could probably do...

```ruby
    def widget
      (@opts[:widget] || "text_field").to_s
    end
```

ericfranz:
Is the smart attribute factory methods, following how we manage job adapters, necessary? What was the rationale?

nickjer:
If it is going to be reusable it doesn't hurt to all third-parties to add their own smart attributes.

nickjer:
I think it looks better with justified pills:

image

compared to tabs:

image

nickjer:
Also having second thoughts about ordering the tabs/pills by Title name. As the more common sub-app may always require two clicks as in the case of Jupyter:

image

It takes me to Jupyter+CUDA whenever I go to the root jupyter url.

ericfranz:
Add the tabs back. Get rid of the <hr> underneath the title:

63925fba-408f-11e7-8045-e176f2006a56


Or for consistency sake (between apps and subapps), get rid of the redundant title, and just use the title in the tab. Always have a tab, even for a single app.

63925fba-408f-11e7-8045-e176f2006a56 copy

ericfranz:
I could be wrong but I think that the tabs offer better affordances than the pills.

nickjer:
Another interface I feel is better looking than tabs:

image

nickjer:
Also pills are more touch screen friendly than the tabs, and you Alan loves his surface

nickjer:
Could even go Google style with underlined tabs/pills:

image

nickjer:
It seems tabs can look funny on smaller screens:

image

nickjer:
Another option is the menu to the left if there are more than one sub-apps:

image

ericfranz:
On smaller screens what do the pills look like

nickjer:
Widescreen:

image

Phone Screen:

image

ericfranz:
Instead of `/var/www/ood/apps/sys/interactive` for the directory... we can make it `/var/www/ood/apps/sys/templates` for "app templates".

1. The manifest can specify a new attribute "renderer: batch_connect_v1" or just "batch_connect" for v1 (and "batch_connect_v2" if we ever go there).
2. The default renderer can be this (if omitted)
3. The url can specify the renderer.

Options for the router:

```
# basically same as we discussed before
/var/ood/apps/sys/templates/desktop
"/pun/sys/dashboard/batch_connect/sys/templates/desktop"

# since we are already parsing it using regex instead of glob in the routes, 
# app token parts could be in different parts of the url
/var/ood/apps/sys/templates/desktop
"/pun/sys/dashboard/templates/batch_connect/sys/desktop"

# for example, this is enough to know we are working with an app template
# and the renderer is batch_connect, and parsing "sys/desktop" or "usr/efranz/desktop"
# into owner and app name is easy (we already do this) so we can then 
# build the app token "sys/templates/desktop"
```

There would no longer be a requirement to specify the URL in the manifest. In fact, we could drop this feature (or deprecate it).

And, we don't have to implement this now, just make sure the routes and the locations of the batch_connect apps are following the convention so when we implement this it will not be an expensive change.

**Benefits:**

It provides a path for adding other app template formats, or newer versions of batch_connect app template formats, while having backwards compatibility with existing batch_connect formats. We could look at adding a batch_workflow app template format, implement support for GenApp, etc.

And of course its a way to have all our templates be in a single directory.

And its low cost as right now all we have to do is deploy our batch_connect apps under a directory called templates and use a sensible URL.

ericfranz:
Two thoughts on rendering.

1. we don't have an example/use case for needing the form.html.erb to be an erb
2. for submit.html.erb, we should consider using Liquid instead. Liquid templates (used in Jekyll) are designed to be "safe for users to edit"

That said, I'd rather not change a thing until we get a test harness in place for the code that loads these app templates into app and form and smart attribute list objects.

nickjer:
> And of course its a way to have all our templates be in a single directory.

I wouldn't recommend mixing Dashboard batch_connect apps with some other Passenger app generator like batch_workflow in the same directory. I feel if we build another Passenger app that generates apps from configuration files, it should look for those apps in a different directory. And that directory should be aptly named to not cause confusion.

It will be easy for the Dashboard batch_connect app to look for all directories in `/var/www/ood/apps/sys/interactive/*` and display them as an app list to the left (see below).

> The manifest can specify a new attribute "renderer: batch_connect_v1" or just "batch_connect" for v1 (and "batch_connect_v2" if we ever go there).

I believe the manifest can be as simple as a url option that points to `/pun/sys/dashboard/batch_connect/sessions` and have the title "Interactive Apps". That is because this will now be a completely self-contained app, assuming we list all the batch connect apps within the "Interactive Apps" app as below:

image

> we don't have an example/use case for needing the form.html.erb to be an erb

I believe you meant `form.yml.erb`, and they can use either `form.yml` or `form.yml.erb` if they'd like. The former won't perform ERB rendering. But I don't want to assume they will never need pre-processing when building the form.

- they can use ERB to build select options dynamically (read a file and build a yaml array from it)
- with a date input field they can use ERB to initialize it with the current date (or a date that is one week from now)
- initialize a textarea field with contents of a file 

> for submit.html.erb, we should consider using Liquid instead

I think you meant `submit.yml.erb`. A couple of reasons to use ERB instead of a simplified templating language:

- we use ERB in the `connection.html.erb` file (we can even provide helper methods in Rails for them to use to build common urls), so they can reuse their ERB knowledge in the submit file
- ERB allows the developer more control, they have full access to the Ruby language if they need to use the standard library (time/date manipulation), do arithmetic (calculate license tokens from cores), store a local variable to be re-used throughout the code (generate a UUID once and use it in multiple places)

ericfranz:
>I wouldn't recommend mixing Dashboard batch_connect apps with some other Passenger app generator like batch_workflow in the same directory.

It would not be another Passenger app; it would be the Dashboard.

>I feel if we build another Passenger app that generates apps from configuration files, it should look for those apps in a different directory. And that directory should be aptly named to not cause confusion.

We were already mixing batch connect apps in the same directory with passenger apps previously. Moving "app templates" to a separate directory from passenger apps is an improvement. But we could get by with either separate app template directories for different template types or store them all in a single one.

ericfranz:
In regards to all of the things you mention in ERB:

1. half of them can be done with liquid templating (basic time/date manipulation, variable assignment, arithmetic)
2. more complex cases can be handled by additions to "smart attributes"

ericfranz:
Let's keep erb for these apps for now.

nickjer:
We need to adjust the style a bit to account for the longer title "Interactive Apps":

image

ericfranz:
We can get rid of home and add those bread crumbs to the dev pages...

ericfranz:
Actually, I'm concerned that even with my fixes, won't Utah have a problem with this? I thought they were using an image in the navbar that was wider than the AweSim image.

ericfranz:
One thought. We have desktop. And we have /local to have overrides. 

We could offer the ability to do something like how keycloak themes work. For keycloak theme you can set parent attribute. And then you only provide files that you are making modifications to.

So an alternative to adding /local underneath an app and adding modifications would be to:

1. create a new app directory
2. specify in the manifest role batch_connect and the parent app being overridden:

    ```yaml
    role: batch_connect
    parent: desktops
    ```

3. add the files you would normally place under local here i.e.

```
efranz@webdev04:/var/www/ood/apps/sys$ tree bc_oakley_desktop/
bc_oakley_desktop/
├── form.yml
├── manifest.yml
├── submit.yml.erb
```

We could optionally deploy the dashboard with the desktops app template included, so any app that sets parent to "desktops" would use this template (unless there was another app named desktops).

ericfranz:
The whole `parent:` idea is cool but a possible feature after initial release.

lib/smart_attributes/attributes/bc_vnc_resolution.rb
```diff
@@ -0,0 +1,39 @@
...
+        (opts[:label] || "Resolution").to_s
+      end
+
+      # Submission hash describing how to submit this attribute
+      # @param fmt [String, nil] formatting of hash
+      # @return [Hash] submission hash
+      def submit(fmt: nil)
+        {
+          batch_connect: {
+            template: "vnc",
```
ericfranz: This attribute should not specify the batch_connect template that is used, just the argument sent to the template. The app should determine the batch connect template that is used.

ericfranz:
We can do this one later after release:

>2. I think we should include the full, final hash, resulting from merging all these submit method call results on attributes and the form, in the session so they are available to the user or developer in the future. Will need to think of a way to present this in the interface.

nickjer:
I think we should merge this in as is so that we can roll it out for next Tuesday.

brianmcmichael:
The "Desktops" page doesn't seem to be inside of a container, or the OOD footer is missing. On my machine, the "Launch" button is flush with the bottom of the page.

footer

nickjer:
There is no footer for the Dashboard, so what you are seeing is what is intended.

app/models/batch_connect/session.rb
```diff
@@ -0,0 +1,337 @@
...
+      cluster.batch_connect_template(template: script_type, **context)
+    end
+
+    # Stage and submit a session from an app and its context
+    # @param app [BatchConnect::App] batch connect app
+    # @param context [BatchConnect::SessionContext] context used for session
+    # @param format [String] format used when rendering template
+    # @return [Boolean] whether saved successfully
+    def save(app:, context:, format: nil)
+      self.id         = SecureRandom.uuid
```
ericfranz: We should probably not use SecureRandom.uuid here but do something that ensures there will never be a conflict. Maybe `self.generate_uniq_id`. It could use SecureRandom.uuid but check if there is a conflict.

Conflict is rare but possible because we have different hosts using the same "database". I'll open a separate issue for this though.

nickjer: I think this is so incredibly rare, that it may never happen within the Earth's lifetime.

nickjer: From the Wikipedia page:

https://en.wikipedia.org/wiki/Universally_unique_identifier

>For example, the number of random version 4 UUIDs which need to be generated in order to have a 50% probability of at least one collision is 2.71 quintillion

nickjer: > This number is equivalent to generating 1 billion UUIDs per second for about 85 years, and a file containing this many UUIDs, at 16 bytes per UUID, would be about 45 exabytes, many times larger than the largest databases currently in existence, which are on the order of hundreds of petabytes.

ericfranz: From the same article

>RFC 4122 advises that "distributed applications generating UUIDs at a variety of hosts must be willing to rely on the random number source at all hosts. If this is not feasible, the namespace variant should be used."

ericfranz: Eh but most people are running one or a few sessions at a time. And thats from one host, not starting multiple sessions. Nevermind.

brianmcmichael: https://tools.ietf.org/html/rfc4122

It's generated based on the system and time and it gets a global lock during generation, so there really shouldn't ever be a conflict.

nickjer: We can rely on the random number generator at all the hosts, it is typically `/dev/urandom`.

nickjer: I think that is meant to warn about security implications. If you aren't using a secure enough random number generator, then a third-party may be able to guess the next UUID you generate.

That isn't an issue for us, because we aren't using it in a security setting.

PR #157, opened 2017-05-23
Lazy load scratch and project paths via ajax

brianmcmichael:
I'm not sure if there's an issue outstanding for this.

Significantly increases page load time by loading additional file explorer paths after the page has loaded.

When Project and Scratch are not responsive, this can slow the entire dashboard. This resolves that by avoiding the system checks on the initial load. The menu is instead silently filled by the user's browser after the page has loaded.


nickjer:
Just my opinion, but if we go this route I recommend adding some sort of loading (busy) indicator. Otherwise it may confuse people when they see the Files dropdown with more options on some occasions and none on others.

A little off topic, but as I am not too familiar with this part of the code, do we check for these file paths only on the front page of the dashboard or on all pages (i.e., setting up an app's db page)?

brianmcmichael:
> if we go this route I recommend adding some sort of loading (busy) indicator

That might be fine, it actually loads pretty quickly when the file system is responsive, so by the time I go to click the "Files" dropdown, the data is already present. My concern would be that most centers probably aren't going to customize favorite_paths, so it may be odd if a user clicks the Files dropdown and sees a 'loading' message, only to have it be replaced by nothing.

> do we check for these file paths only on the front page of the dashboard or on all pages (i.e., setting up an app's db page)

This is a good question, we don't have app sharing deployed in the base OOD deployment, so it wasn't tested there. I can test in an AweSim implementation and find out the behavior.

nickjer:
> My concern would be that most centers probably aren't going to customize favorite_paths

You can add a class to the Files navbar that tells it whether it needs to be loaded with elements from Ajax. And only add that class if the favorites_path isn't empty.

> we don't have app sharing deployed in the base OOD deployment

The dashboard will always link to itself and look for a `bin/setup-production` no matter what. For example this is the SystemStatus app link in the dashboard menu:

```html
<a href="/pun/sys/dashboard/apps/show/systemstatus" target="_blank"><i class="fa fa-tachometer app-icon" title="FontAwesome icon specified: tachometer"></i> System Status</a>
```

So it will always route the user to `/pun/sys/dashboard/apps/show/<app>`.

ericfranz:
The alternative https://github.com/OSC/ood-dashboard/pull/205 (see comments/discussion on options) or a more robust circuit breaker solution is the right approach.

Though this PR has great code examples using the js-routes gem, looks like an interesting option for using AJAX to pull rendering of HTML erb partials. The code is really clean. Something to consider in the future. I'll close this PR now though.

PR #166, opened 2017-06-12
Ihcp interactive dir

ericfranz:


ericfranz:
1. Makes it more complex for app installer scripts (?)
2. Google Analytics (assumes apps under 1 directory) and what is an app vs a route underneath the app
3. Rails application template and its dotenv files (find the app name by just looking at the directory name)

Could avoid Passenger apps being deployed in subdirectories (just plugins).

ericfranz:
No longer doing.

PR #169, opened 2017-06-15
Show all subapps interactive apps lists

ericfranz:
A possible fix. It is currently brittle though. Calling `BatchConnect::App#title` may result in `AppNotFound` exception.

screen shot 2017-06-15 at 3 29 32 pm

Also if the height of a window is 768px, what we often do for screenshots, the result is Jupyter app gets cut off:

screen shot 2017-06-15 at 3 31 48 pm

In this case, it might be preferable anyways to have subapps expand only when selected, by doing this:

screen shot 2017-06-15 at 2 36 17 pm

### DIFF
--- a/app/controllers/batch_connect/session_contexts_controller.rb
+++ b/app/controllers/batch_connect/session_contexts_controller.rb
@@ -2,16 +2,9 @@ class BatchConnect::SessionContextsController < ApplicationController
   # GET /batch_connect/<app_token>/session_contexts/new
   def new
     set_app
-    set_sub_apps
     set_render_format
     set_session_context
     set_apps
-
-    # Redirect to first valid sub app otherwise first invalid sub app
-    unless @sub_apps.include? @app
-      sub_app_token = ( @sub_apps.select(&:valid?).first || @sub_apps.first ).token
-      redirect_to new_batch_connect_session_context_url(token: sub_app_token)
-    end
 
     if @app.valid?
       # Read in context from cache file
@@ -31,7 +24,6 @@ class BatchConnect::SessionContextsController < ApplicationController
   # POST /batch_connect/<app_token>/session_contexts.json
   def create
     set_app
-    set_sub_apps
     set_render_format
     set_session_context
     set_apps
@@ -62,11 +54,6 @@ class BatchConnect::SessionContextsController < ApplicationController
     def set_apps
       # get lists of apps
       @apps = sys_app_groups.select(&:has_batch_connect_apps?)
-    end
-
-    # Set the list of sub apps
-    def set_sub_apps
-      @sub_apps = @app.sub_app_list
     end
 
     # Set the session context from the app
--- a/app/controllers/batch_connect/session_contexts_controller.rb
+++ b/app/controllers/batch_connect/session_contexts_controller.rb
@@ -15,9 +15,6 @@ class BatchConnect::SessionContextsController < ApplicationController
         #{@app.validation_reason} Please contact support if you see this message
       EOT
     end
-  rescue BatchConnect::App::AppNotFound => e
-    flash.now[:alert] = e.message
-    @app = nil
   end
 
   # POST /batch_connect/<app_token>/session_contexts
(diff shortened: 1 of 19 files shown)


ericfranz:
Yes. Make one of the subapps poorly defined and you get errors such as `Psych::SyntaxError`. Could also result in `AppNotFound`.

ericfranz:
This implementation is a bad one. We will hold off on this.

ericfranz:
@nickjer I need to fix another unit test but otherwise its ready for review

app/models/batch_connect/app.rb
```diff
@@ -93,13 +99,19 @@ def cluster
     # Whether this is a valid app the user can use
     # @return [Boolean] whether valid app
     def valid?
+      return false if @validation_reason
```
nickjer: This won't be set until `#form_config` has been called. So something like `App.new(router: router).valid?` would skip this line, but it would still return `false` when it called `cluster` in the line below with the proper validation reason.

Although the proper error is still captured, it is confusing which line of code was responsible for it.

So one suggestion would be...

```ruby
def valid?
  !form_config.empty? && cluster && cluster.job_allow?
end
```

This makes it slightly clearer what we are testing for.

app/controllers/batch_connect/session_contexts_controller.rb
```diff
@@ -22,16 +15,12 @@ def new
         #{@app.validation_reason} Please contact support if you see this message
       EOT
     end
-  rescue BatchConnect::App::AppNotFound => e
```
nickjer: Since you removed this, you will need to add periods `.` to the end of all errors. So the error messages have proper-ish grammar.

Currently they will see something like this...

```
This app does not supply a form file under the directory '/path/to/app' Please contact support if you see this message
```

We should probably put a period between the error message and the "Please contact...".

ericfranz: Hmm... I guess normal exception messages don't add periods, so it would need to be something ugly like:

```ruby
rescue AppNotFound => e
  @validation_reason = "#{e.message}."
```

Or we just remove "Please contact support if you see this message". We actually don't display that anywhere else an error occurs in our apps.

ericfranz: Actually, these errors will only appear if the batch connect app is incorrectly built. An incorrectly built batch connect app should not be deployed to production. Unless the user is trying to access an app that does not exist.

ericfranz: So I'll just remove "Please contact support if you see this message". If you object lets discuss offline.

nickjer:
Tested:

- sys/desktop/form.yml missing

  screenshot from 2017-07-14 11-08-47
  screenshot from 2017-07-14 11-05-46

- sys/desktop/form.yml invalid (bad syntax in file)

  screenshot from 2017-07-14 11-08-01
  screenshot from 2017-07-14 11-07-14

- sys/desktop/form.yml has no cluster (`cluster: null` or no defined `cluster`)

  screenshot from 2017-07-14 11-11-00

  Sub apps work as expected.

- sys/desktop/form.yml has invalid cluster (`cluster: bob`)

  screenshot from 2017-07-14 11-13-01

  Sub apps work as expected.

- sys/desktop/form.yml has 600 permission (I can't read it)

  screenshot from 2017-07-14 11-14-23
  screenshot from 2017-07-14 11-14-45

- I visit the URL of a sub-app that doesn't exist for a given app (https://apps-dev.awesim.org/pun/dev/ood-dashboard/batch_connect/sys/bc_desktop_v2/bob/session_contexts/new)

  screenshot from 2017-07-14 11-18-45

- Invalid sub-app yaml config (messed up the Oakley desktop config)

  Verify it doesn't show up in menu, but other sub-apps do:

  screenshot from 2017-07-14 11-21-04
  screenshot from 2017-07-14 11-21-37

- sys/desktop/local has 700 permissions (I can't go into the directory)

  screenshot from 2017-07-14 11-24-49

  Uh oh!

- I delete sys/desktop/local

  screenshot from 2017-07-14 11-25-35

nickjer:
By the way @ericfranz, I am fine with this PR if you fix the above crash. You can merge it in after that.

nickjer:
Oh yeah, and don't forget to update the `CHANGELOG.md` when you are done.

ericfranz:
Recent commits addressed the crash. Please review. I'll also fix the final failing test and update the CHANGELOG.

PR #178, opened 2017-06-27
Fix cache file extension

nickjer:


nickjer:
This is patched in #182

PR #180, opened 2017-07-02
Add iHPC rake task

nickjer:
Provides the capability to launch an iHPC session from the command line using a Rake task. See modified `README.md` for more details.

### DIFF
--- a/lib/tasks/batch_connect.rake
+++ b/lib/tasks/batch_connect.rake
@@ -4,13 +4,7 @@ namespace :batch_connect do
     # Read in user settings
     app = ENV["BC_APP_TOKEN"] || abort("Missing environment variable BC_APP_TOKEN")
     fmt = ENV["BC_RENDER_FORMAT"]
-    ctx = if !$stdin.tty?
-            $stdin.read
-          elsif ENV["BC_SESSION_CONTEXT"]
-            File.read(ENV["BC_SESSION_CONTEXT"])
-          else
-            "{}"
-          end
+    ctx = $stdin.read
 
     # Initialize objects
     app   = BatchConnect::App.from_token app
--- a/lib/tasks/batch_connect.rake
+++ b/lib/tasks/batch_connect.rake
@@ -20,6 +20,12 @@ namespace :batch_connect do
 
     # Generate new session
     session = BatchConnect::Session.new
-    session.save(app: app, context: session_ctx, format: fmt)
+    unless session.save(app: app, context: session_ctx, format: fmt)
+      error_msg = "Failed to launch session with following errors:\n\n"
+      session.errors.full_messages.each do |error|
+        error_msg += "  - #{error.gsub(/\n/, "\n    ")}\n\n"
+      end
+      abort error_msg
+    end
   end
 end
--- a/lib/tasks/batch_connect.rake
+++ b/lib/tasks/batch_connect.rake
@@ -3,14 +3,20 @@ namespace :batch_connect do
   task new_session: :environment do
     # Read in user settings
     app = ENV["BC_APP_TOKEN"] || abort("Missing environment variable BC_APP_TOKEN")
-    ctx = ENV["BC_SESSION_CONTEXT"]
     fmt = ENV["BC_RENDER_FORMAT"]
+    ctx = if !$stdin.tty?
+            $stdin.read
+          elsif ENV["BC_SESSION_CONTEXT"]
+            File.read(ENV["BC_SESSION_CONTEXT"])
+          else
+            "{}"
+          end
 
     # Initialize objects
     app   = BatchConnect::App.from_token app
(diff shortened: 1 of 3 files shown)


ericfranz:
@nickjer @basilgohar @brianmcmichael 

This is a good example of why we should consider moving from "GitHub Flow" to "Git Flow" branching model for the core ondemand apps. Here, we have a significant change to the README which will display on the front page as soon as we merge into master. However, the latest commit to master has not yet been tagged as a release, tested, and confidently deployed at OSC. Until we do this, there is a period of time where someone outside of OSC could navigate to the dashboard repo, read the README, see the documentation on the CLI, try to run it, and it would fail because it doesn't exist yet.

brianmcmichael:
For clarification, you're talking about something like [this](http://nvie.com/posts/a-successful-git-branching-model/), right?

git-model

lib/tasks/batch_connect.rake
```diff
@@ -0,0 +1,31 @@
+namespace :batch_connect do
+  desc "Generate new batch connect session"
+  task new_session: :environment do
+    # Read in user settings
+    app = ENV["BC_APP_TOKEN"] || abort("Missing environment variable BC_APP_TOKEN")
+    fmt = ENV["BC_RENDER_FORMAT"]
+    ctx = if !$stdin.tty?
```
ericfranz: Why not just:

```ruby
ctx = $stdin.read
ctx = "{}" if ctx.strip.empty?
```

Is there a reason to support multiple ways to run this from the command line? i.e. if I can pipe cat mycontext.json to the command then we don't need BC_SESSION_CONTEXT.

I guess the answer depends on what the reason was for adding this rake task in the first place... (easy testing on other systems?)

nickjer: That would cause the rake task to always try to read from stdin. If a user just wanted to submit a job with default values, he'd have to remember to pipe in `"{}"` or the rake task will just hang waiting for input.

ericfranz:
@brianmcmichael yes. gitflow adds the three branches in the middle (develop, release branches, hotfixes) to feature branches and master (GitHub Flow). We'll move discussion out of this issue though.

lib/tasks/batch_connect.rake
nickjer: > I guess the answer depends on what the reason was for adding this rake task in the first place... 

I thought the reason was so other apps or utilities could launch iHPC sessions on behalf of the user. So reading from stdin may not always be the easiest option depending on the language chosen by the developer.

nickjer: Actually I would need to test whether the solution I provided would still skip the stdin reading if it was launched from a process without a tty.

nickjer: In tools that accept a file or stdin, if the filename provided is `-` then it reads from stdin. Unfortunately rake isn't the easiest to supply arguments. So another solution would be to read from stdin if `BC_SESSION_CONTEXT=-`.

nickjer: An example node.js web app that launches this rake task using stdin:

```javascript
const express = require('express')
const app = express()

const cp = require('child_process')
const extend = require('util')._extend

app.get(process.env.PASSENGER_BASE_URI + '/', function (req, res) {
  var body = ''
  var rake = cp.spawn('bin/rake', ['batch_connect:new_session'], {
    cwd: '/users/appl/jnicklas/ondemand/dev/ood-dashboard',
    env: extend(process.env, {
      PWD: '/users/appl/jnicklas/ondemand/dev/ood-dashboard',
      RAILS_ENV: 'development',
      RAILS_RELATIVE_URL_ROOT: '/pun/dev/ood-dashboard',
      BC_APP_TOKEN: 'sys/bc_desktop_v2/owens'
    })
  })

  rake.stdin.write('{}')
  rake.stdin.end()

  rake.stdout.on('data', function (data) {
    body += data
  })
  rake.stderr.on('data', function (data) {
    body += data
  })
  rake.on('close', function (code) {
    res.send(body)
  })
  rake.on('error', function (err) {
    res.send(err)
  })
})

app.listen(3000, function () {
  console.log('Example app listening on port 3000!')
})
```

nickjer: Unfortunately, the Node.js app has no tty, so there is no way to implicitly determine if STDIN should be used. So we either fully use STDIN, so launching a default Owens Desktop would look like:

```sh
echo "{}" | bin/rake batch_connect:new_session RAILS_RELATIVE_URL_ROOT=dev/ood-dashboard BC_APP_TOKEN=sys/bc_desktop_v2/owens
```

or we provide an explicit way to turn on or off STDIN checking.

ericfranz: Seems like easiest is to do `BC_SESSION_CONTEXT=-`

PR #182, opened 2017-07-03
Fixed batch_connect dataroot

### LINKED ISSUE #181: Namespace batch_connect dataroot
As @ericfranz pointed out we will need to namespace the dataroot used by the batch_connect namespaced code. This can affect active sessions (queued and running) so a in-situ migration may need to be applied.

nickjer:
Fixes #181

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,6 +43,7 @@ tags
 /public/assets
 .env.local
 /config/initializers/ood.rb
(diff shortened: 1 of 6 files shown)


nickjer:
If there are jobs it moves the relevant directories underneath `batch_connect/` and makes symlinks at the old locations.

If there are no jobs it just deletes the relevant directories if they exist.

Issue #189, opened 2017-07-10
Browser requirements warning appears only on main page


nickjer:
This is just a question and not an issue. I noticed that we display the browser requirements warning on the main Dashboard index page as such:

![screenshot from 2017-07-10 14-22-07](https://user-images.githubusercontent.com/4260509/28033159-2e499c22-657b-11e7-8d1a-114b7e009c00.png)

But we do not display it on any other Dashboard page. For example it is not displayed on the iHPC pages or error pages.

Is this intended behavior? Or do we want to display this on every Dashboard page (i.e., make it part of the layout)?

ericfranz:
Lets not worry about this right now.

PR #194, opened 2017-07-11
Move the alert flash

### LINKED ISSUE #191: Browser requirements not showing up on AweSim
I see this on OnDemand when using an old browser:

screenshot from 2017-07-10 14-22-07

But I don't see it on AweSim:

screenshot from 2017-07-10 15-22-37

brianmcmichael:
Resolves #191 and #189 

Move the alert flash from the dashboard controller to the application controller to show on all dashboards.

The AweSim dashboard relies on a separate controller than the default OOD index view. I've moved the detection and flash setter to the ApplicationController where it will apply to all of the views, including the AweSim dashboard.

browser_warn


nickjer:
This breaks alert messages meant to be displayed on redirects. For instance, it breaks the following line of code:

https://github.com/OSC/ood-dashboard/blob/09fdd6e2ba951cf61afd7eeac2a50a0e40234572/app/controllers/batch_connect/sessions_controller.rb#L29

You can reproduce this with the following example.

If I create a session in the old browser I see:

screenshot from 2017-07-14 11-50-11

But when I go to delete this session and `qdel` fails for some reason or another I see:

screenshot from 2017-07-14 11-50-42

nickjer:
One solution is to have this message treated special in the layout page. Similar to:

https://github.com/OSC/ood-dashboard/blob/09fdd6e2ba951cf61afd7eeac2a50a0e40234572/app/views/layouts/application.html.erb#L46-L60

brianmcmichael:
An alternative solution is to treat the `alert` flash as an array rather than a string, then append new messages to it. In the view, iterate over them and produce an alert message for each element.

brianmcmichael:
So, the simplest solution seemed to be to hijack @nickjer's Safari warning and use the same key for each, since they are all browser errors, it's ok in this case if one overwrites another. This avoids the conflict with traditional actions on `flash[:alert]` and keeps the desired functionality.

See specifics at: https://github.com/OSC/ood-dashboard/commit/8793e7eb75b4365b70e65769a046a4ddcc2cdac3

nickjer:
I feel this code is becoming more-and-more similar to just inlining the conditional warning messages directly into the layout, albeit through a more round-about way with before_filters.

Just an opinion so feel free to reject it, but we could accomplish this as well with a partial that defines browser requirements. This would avoid the scary use of a `before_filter` as well as avoid using the `view_context` within the controller.

Otherwise I am fine with the code changes other than my comment above about a possible missing `view_context` on one line of code.

ericfranz:
This is my fault for encouraging the previous approach. 

Lets make it a partial.

Before it was just set on GET dashboard/index now its being set on every request. If it is a partial in the application layout, or just in the application layout, then it will only be done in requests that render html using the layout.

brianmcmichael:
Converted this to a partial and render in the application layout.

Issue #29, opened 2016-10-12
Checking for project space path existence when unavailable causes very slow request


ericfranz:
The request to load the dashboard when GFPS is unavailable is very slow. Here, it loaded in 1.7 minutes:

![screen shot 2016-10-12 at 5 12 27 pm](https://cloud.githubusercontent.com/assets/512333/19327866/1b18f15e-909f-11e6-9572-647c2ffb8dec.png)

This is because I have two groups that start with `P` and so the dashboard looks for the existence of both `/fs/project/PZS0562` and `/fs/project/PZS0530`.

We probably should time-out when trying to determine if these don't exist, and perhaps let the user know that we timed-out when trying to determine the existence of these directories.

ericfranz:
A note: when the process hangs like this, sometimes not even SIGKILL will work. Here, I `kill -9 49811` but it is still there.

![screen shot 2016-12-27 at 2 36 56 pm](https://cloud.githubusercontent.com/assets/512333/21507677/7bf9b118-cc48-11e6-90cb-08b1c4b521b2.png)

nickjer:
Is the `kill -9 49811` off the screen? Because in the left-window I see you drop out of the wait in the current process, but you don't mess with the actual process doing the file list.

You might need to generate a thread with `sleep` and if it reaches the end of the `sleep` you `Process.kill(9, 46811)`, if the `waitpid` returns results before then you terminate the `sleep` thread.

ericfranz:
Yes, `kill -9 49811` is off the screen.

ericfranz:
Another option that might work:

https://github.com/AweSim-OSC/vncsim/blob/760035052f6ca7d2e1ce4e79033370c86fc0fd5f/app/models/session.rb#L267-L271

The point of the `Dir.open(outdir).close` was to force nfs to update the files in the directory. We could use a similar method. Fork and do the Dir.open(path).close and just check to see if the process hangs; if it doesn't you might be good to go.

ericfranz:
In production for a period of time all of the OnDemand and AweSim dashboards became unresponsive to the point of having a Proxy Error because of this. The dashboard was completely unresponsive and SUG is going on.

![screen shot 2017-04-06 at 11 54 54 am](https://cloud.githubusercontent.com/assets/512333/24763672/1174567e-1ac0-11e7-826d-575b8ae7d8a5.png)

This is the first time I've observed this problem outside of downtimes where hiccups are expected.

[...]

PR #205, opened 2017-07-18
Add a timeout to the path check to speed up load on hanging filesystem

brianmcmichael:
Alternative to https://github.com/OSC/ood-dashboard/pull/157

This adds a timeout around the filesystem check. Currently at 3 seconds, but that's arbitrary.

The negatives of this solution are that if the system is slow it won't show the additional paths, even if they may be available, and also that the page will still hang for a few seconds (but not 30) when the paths are unresponsive.

The positive is obviously that it's simple and completely server-side.

### DIFF
--- a/app/apps/ood_files_app.rb
+++ b/app/apps/ood_files_app.rb
@@ -19,7 +19,17 @@ class OodFilesApp
   # returns an array of other paths provided as shortcuts to the user
   def favorite_paths
     @favorite_paths ||= candidate_favorite_paths.select {|p|
-      p.directory? && p.readable? && p.executable?
+      path_valid p
     }
+  end
+
+  private
+
+  def path_valid(path)
+    Timeout.timeout(3) {
+      path.directory? && path.readable? && path.executable?
+    }
+  rescue Timeout::Error => e
+    false
   end
 end


ericfranz:
This won't work. Already tried it. The process will still hang for 30 seconds to a minute.

brianmcmichael:
It does work. Add a `sleep 10` to the timeout block and try it.

ericfranz:
`sleep 10` is not the same thing as a stale file handle or something worse

ericfranz:
Here is the original issue https://github.com/OSC/ood-dashboard/issues/29.

Actually, for some cases just doing something like this might work:

```diff
  def favorite_paths
+   cmd = candidate_favorite_paths.map { |p| "ls #{p}"}.join("; ")
+   Timeout::timeout(5) { `#{cmd}` }
+   # Timeout::timeout(5) { `ls /fs/project; ls /fs/scratch` }
+
    @favorite_paths ||= candidate_favorite_paths.select {|p|
      p.directory? && p.readable? && p.executable?
    }
+ rescue Timeout::Error
+   @favorite_paths = []
+ end
```

Of course we would need to use open3 and the correct commands to properly escape the files.

I don't know why I didn't think of it before. But thats based on the idea that this would be true (which isn't always the case):

>However, Timeout::timeout will correctly cancel a Process.wait call if the child process that Process.wait is waiting on is in an uninterruptible sleep state:
>
>```
>irb(main):012:0> Timeout::timeout(5) { `ls /fs/project` }
>Timeout::Error: execution expired
>        from (irb):12:in ``'
>        from (irb):12:in `block in irb_binding'
>        from /opt/rh/rh-ruby22/root/usr/share/ruby/timeout.rb:89:in `block in timeout'
>        from /opt/rh/rh-ruby22/root/usr/share/ruby/timeout.rb:34:in `block in catch'
>        from /opt/rh/rh-ruby22/root/usr/share/ruby/timeout.rb:34:in `catch'
>        from /opt/rh/rh-ruby22/root/usr/share/ruby/timeout.rb:34:in `catch'
>        from /opt/rh/rh-ruby22/root/usr/share/ruby/timeout.rb:104:in `timeout'
>        from (irb):12
>        from /opt/rh/rh-ruby22/root/usr/bin/irb:11:in `<main>'
>```
>

nickjer:
You will definitely need to fork off the directory checking as it does completely freeze up the thread it is run on and you are unable to interrupt it.

One idea is:

```ruby
require 'timeout'

begin
  Timeout.timeout(10) do
    # Fork off your checker
    @pipe = IO.popen("....")
    Process.wait @pipe.pid
  end
rescue Timeout::Error
  Process.kill 9, @pipe.pid
  Process.waite @pipe.pid # zombies are bad
end

ericfranz:
The original issue https://github.com/OSC/ood-dashboard/issues/29 has the suggested solution, a circuit breaker, called out. This code is still useful as a reference. https://github.com/OSC/ood-dashboard/issues/29 will have a reference to this so we can close this now.

PR #207, opened 2017-07-20
Fix view-only conn with noVNC and latest turbovnc

### LINKED ISSUE #206: TurboVNC 2.1.1 fixes resizing issue but introduces bug in "View Only"
TurboVNC 2.1.1 fixes a security issue where the "View Only" connections can auto-resize the VNC session when they shouldn't have the power to do so.

Unfortunately noVNC fails to display changes in a VNC session to a "View Only" connection if it is set to `resize=remote`. It is recommended we change this to `resize=downscale` instead for "View Only" connections.

nickjer:
Fixes #206

### DIFF
--- a/app/helpers/batch_connect/sessions_helper.rb
+++ b/app/helpers/batch_connect/sessions_helper.rb
@@ -131,7 +131,7 @@ module BatchConnect::SessionsHelper
 
   def novnc_link(connect, view_only: false)
     version = browser.safari? ? "0.6.2" : "55b459b"
-    asset_path("noVNC-#{version}/vnc.html?autoconnect=true&password=#{view_only ? connect.spassword : connect.password}&path=rnode/#{connect.host}/#{connect.websocket}/websockify&resize=remote", skip_pipeline: true)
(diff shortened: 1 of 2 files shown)


app/helpers/batch_connect/sessions_helper.rb
```diff
@@ -131,7 +131,7 @@ def delete(session)
 
   def novnc_link(connect, view_only: false)
     version = browser.safari? ? "0.6.2" : "55b459b"
-    asset_path("noVNC-#{version}/vnc.html?autoconnect=true&password=#{view_only ? connect.spassword : connect.password}&path=rnode/#{connect.host}/#{connect.websocket}/websockify&resize=remote", skip_pipeline: true)
+    asset_path("noVNC-#{version}/vnc.html?autoconnect=true&password=#{view_only ? connect.spassword : connect.password}&path=rnode/#{connect.host}/#{connect.websocket}/websockify&resize=#{view_only ? "downscale" : "remote"}", skip_pipeline: true)
```
ericfranz: this is getting a little dense. maybe remove the if/else inside interpolation:

```ruby
password = view_only ? connect.spassword : connect.password
resize = view_only ? "downscale" : "remote"

# use #{password} and #{resize} in path

PR #208, opened 2017-07-21
Fixes safari issue with latest novnc

nickjer:
Uses latest version of noVNC across all browsers (including Safari). Allows for a more consistent UI across browsers.


nickjer:
This can be tested on both:

https://ondemand3-dev.hpc.osc.edu

and

https://apps-dev.awesim.org

using the Safari browser.

PR #211, opened 2017-08-09
Add support for ERB rendering of BC app template files

### LINKED ISSUE #165: Provide iHPC developer access to form submission values, resulting job script and Script object, etc.
To debug job submission errors, the form values that the user submitted, along with the resulting Script object used, must be accessible to the user. Prefer putting Script attributes besides content in one file (yaml?) and Script#content in another (i.e. a bash script) so its easier to read.

Original description:

One suggestion is:

> I think we should include the full, final hash, resulting from merging all these submit method call results on attributes and the form, in the session so they are available to the user or developer in the future. Will need to think of a way to present this in the interface.

### LINKED ISSUE #179: iHPC apps should support rendering erb files in the template directory
Take https://github.com/OSC/bc_osc_jupyter/tree/01f68ce8253a86a47842c01a7e6d14ec86b1a9d0 as an example.

Lets say I wanted to add a checkbox to the webform "Use rmpi" and when checked the script.sh will do `module load openmpi`.

It appears that the only option I have to implement this is currently to do this.

1. modify form.yml.erb

```diff
#...
attributes:
+  user_rmpi: { widget: checkbox, label: "Use rmpi" }
  #...
form:
  - bc_num_hours
  - bc_num_slots
  - node_type
  - bc_account
  - bc_email_on_started
+ - user_rmpi
```

2. modify submit.yml.erb

```diff
---
script:
  job_environment:
    PYTHON_MODULE: "python/3.5"
    <%- if node_type.include? "gpus" -%>
    CUDA_MODULE: "cuda/8.0.44"
    <%- end -%>
+   <%= "OPENMPI_MODULE: 'openmpi'" if use_rmpi %>
```

3. modify template/script.sh

```diff
 # Load the require modules
 module load ${PYTHON_MODULE}
 [[ ${CUDA_MODULE} ]] && module load ${CUDA_MODULE}
+[[ ${OPENMPI_MODULE} ]] && module load ${OPENMPI_MODULE}
```

The use of the intermediate env variable in order to modify the script seems limiting. With our app kit apps the entire job template can contain template files which are rendered. If we supported this same approach then I could skip the extra USE_RMPI env var:

2. modify template/script.sh.erb

```diff
 # Load the require modules
 module load ${PYTHON_MODULE}
 [[ ${CUDA_MODULE} ]] && module load ${CUDA_MODULE}
+<%= "module load openmpi" if use_rmpi %>
```

That said, this is currently the only example in iHPC apps where we are using an env var to compensate for the lack of templating in the job scripts. So we can hold off until we have more use cases demonstrated here.

nickjer:
Fixes #179
Fixes #165

### DIFF
--- a/app/models/batch_connect/session.rb
+++ b/app/models/batch_connect/session.rb
@@ -3,6 +3,16 @@ module BatchConnect
     include ActiveModel::Model
     include ActiveModel::Serializers::JSON
 
+    # This class describes the object that is bound to the ERB template file
+    # when it is rendered
+    TemplateBinding = Struct.new(:session, :context) do
+      # Get the binding for this object
+      # @return [Binding] this object's binding
+      def get_binding
+        binding()
+      end
+    end
+
     # Unique identifier that describes this session
     # @return [String] session id
     attr_accessor :id
@@ -126,20 +136,34 @@ module BatchConnect
       self.view       = app.session_view
       self.created_at = Time.now.to_i
 
-      stage(root: app.root.join("template")) &&
+      stage(app.root.join("template"), context: context) &&
         submit(app.submit_opts(context, fmt: format))
     end
 
     # Stage the app's job template to user's staging directory
     # @param root [#to_s] root directory that gets staged
+    # @param context [Object] context available to post-processed staged files
     # @return [Boolean] whether staged successfully
-    def stage(root:)
+    def stage(root, context: nil)
+      # Sync the template files over
       oe, s = Open3.capture2e("rsync -av --delete #{root}/ #{staged_root}")
       unless s.success?
         errors.add(:stage, oe)
         Rails.logger.error(oe)
+        return false
       end
-      s.success?
+
+      # Post process the template files
+      template_binding = TemplateBinding.new(self, context)
+      template_files.each do |template_file|
+        rendered_file = template_file.sub_ext("")
+        template = template_file.read
+        rendered = ERB.new(template, nil, "-").result(template_binding.get_binding)
+        template_file.rename rendered_file
+        rendered_file.write(rendered)
+      end
+
+      true
     end
 
     # Submit session's script to the cluster and record job id to database
@@ -150,7 +174,7 @@ module BatchConnect
       db_file.write(to_json)
       true
     rescue ClusterNotFound, AdapterNotAllowed, OodCore::JobAdapterError => e
-      output_root.rmtree
+      staged_root.rmtree
       errors.add(:submit, e.message)
       Rails.logger.error(e.message)
       false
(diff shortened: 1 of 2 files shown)


app/models/batch_connect/session.rb
```diff
@@ -126,20 +136,34 @@ def save(app:, context:, format: nil)
...
         errors.add(:stage, oe)
         Rails.logger.error(oe)
+        return false
       end
-      s.success?
+
+      # Post process the template files
+      template_binding = TemplateBinding.new(self, context)
+      template_files.each do |template_file|
+        rendered_file = template_file.sub_ext("")
```
ericfranz: `rendered_filename`

app/models/batch_connect/session.rb
```diff
@@ -126,20 +136,34 @@ def save(app:, context:, format: nil)
...
+      # Sync the template files over
       oe, s = Open3.capture2e("rsync -av --delete #{root}/ #{staged_root}")
       unless s.success?
         errors.add(:stage, oe)
         Rails.logger.error(oe)
+        return false
       end
-      s.success?
+
+      # Post process the template files
```
ericfranz: Instead of a comment `# Post process template files` maybe more clearly `# Render all template files using ERB`?

ericfranz:
Looks good. Missing tests. 

In the future we should consider replacing custom staging code with updated app kit that we release. That would have tests.

nickjer:
I have a concern with the way YAML handles `nil`. It threw me for a loop for awhile thinking the code the was broken and was submitting my job with empty strings.

Here is an example:

```ruby
puts({a: nil, b: "hi", c: ""}.to_yaml)
#=> ---
#=> :a: 
#=> :b: hi
#=> :c: ''
```

Notice that for the value `a` it is not immediately obvious this is set to `nil`.

But JSON makes this far more obvious:

```ruby
puts JSON.pretty_generate({a: nil, b: "hi", c: ""})
#=> {
#=>   "a": null,
#=>   "b": "hi",
#=>   "c": ""
#=> }
```

app/models/batch_connect/session.rb
```diff
@@ -182,14 +205,22 @@ def script(opts = {})
...
       end
 
-      OodCore::Job::Script.new(
-        {
-          content: script_content(bc_opts),
-          job_name: job_name,
-          workdir: output_root,
-          output_path: output_file
-        }.merge script_opts
+      script_opts = {
```
ericfranz: This seems like the wrong method to place this in. This method used to be mostly functional and had one purpose: `Session#script` is defined as a method to build an `OodCore::Job::Script` given a hash of options. Its only state is the `cluster.job_config[:adapter]`, which arguably should be an argument as well.

Now, when ever you get the `script` for the corresponding options hash, the data is written to the staged directory.

Probably more appropriate would be a separate method that accepts `OodCore::Job::Script` instance as an argument and writes the data to the files.

nickjer: > Its only state is the `cluster.job_config[:adapter]`, which arguably should be an argument as well.

It also has `job_name`, `staged_root`, and `output_file` as state.

> Probably more appropriate would be a separate method that accepts `OodCore::Job::Script` instance as an argument and writes the data to the files.

The `OodCore::Job::Script` isn't serializable by design. So this wouldn't work, unless we refactored it to be serializable.

Another solution would be to create a new method called `Session#script_opts` that generates the initialization hash that is then fed to `OodCore::Job::Script#initialize`.

ericfranz:
Okay. In regards to the CHANGELOG prefixing with (Batch Connect) we need to be sure to call out  and define "Batch Connect plugin" in the documentation.

Issue #168, opened 2017-06-13
Icons appear when full text is appropriate


nickjer:
When navbar shrinks enough that we get the full dropdown menu, some of the navbar options display just icons instead of descriptive text:

![image](https://user-images.githubusercontent.com/4260509/27085709-14e1512a-501e-11e7-888e-e7c5371ca6cb.png)

nickjer:
I also noticed that the user name is offset to the left.

PR #221, opened 2017-08-10
Fix responsive nav issues

### LINKED ISSUE #213: Fix responsive concerns with develop dropdown by switching to an absolutely positioned navbar
We could stop using fixed menu at the top. Use an absolutely positioned menu instead - just like the navbar at http://getbootstrap.com/components/ page at the very top. If you scroll away, you don't see the menu.

Because the main problem with the nav bar taking up two rows is that it is fixed and the padding doesn't auto-update to accommodate for this. Plus, a fixed menu at the top is bad for smaller screens anyways.

---

#### Original description and exploration below:

This was partially addressed before. The navbar does collapse at 768px width.

Bootstrap suggests three options http://getbootstrap.com/components/#navbar

> a. Reduce the amount or width of navbar items.
> b. Hide certain navbar items at certain screen sizes using responsive utility classes.
> c. Change the point at which your navbar switches between collapsed and horizontal mode. Customize the `@grid-float-breakpoint` variable or add your own media query.

The default value for `@grid-float-breakpoint1` is `768px`. I tried a. and b. above but there are still problems:

screen 2017-08-10 at 11 17 27 am

One suggestion was to:

1. Move Develop dropdown to far right next to to the help
2. Show the full text "Develop" and hide the text when its small
3. Change the `@grid-float-breakpoint1` to be `900px` instead of `768px`.

The result would be making the browser smaller than this point will collapse instead of putting a second row:

screen 2017-08-10 at 11 16 18 am

But that only works if you have a username like "efranz". But it doesn't prevent something like this with a long username:

screen 2017-08-10 at 11 15 14 am

ericfranz:
Fixes #213 and #168 

Navbar is now absolutely positioned to top. This is better for mobile and web.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
+- Navbar is more responsive with develop menu consistent with other dropdowns
+- Fixed viewport when viewing on mobile devices
+
+### Changed
+
+- Nav bar is absolutely positioned for responsive issues. The result is that
+  scrolling will shift the navbar out of the viewport.
+
 ## [1.15.2] - 2017-07-24
 
 ### Fixed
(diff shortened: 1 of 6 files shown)


nickjer:
You should add links to the issues in the CHANGELOG.

ericfranz:
Done

PR #225, opened 2017-08-12
Add apps to develop nav

ericfranz:
Show up to 6 recent apps:

screen 2017-08-11 at 8 10 48 pm

Two problems with this solution:

1. (minor) when going to the sandbox apps index page `ProductsController#index` the list of apps is built twice
2. (major) OodApp#modified_at just gets the modified date of the directory which does not change when you do any of these actions:
   1. change a file in that directory
   2. touch tmp/restart.txt to restart the app
   3. launch the app
   4. go to the details page of the app

For this reason, this feature would only become useful if we used another value besides the directory modified date. This is a similar problem of sorting by modified date in the apps list. Also, it might be preferable, instead, for the user to **star** specific apps as "favorites" and then for those to appear in the menu.

### DIFF
--- a/app/apps/ood_app.rb
+++ b/app/apps/ood_app.rb
@@ -23,6 +23,18 @@ class OodApp
     manifest.name.empty? ? name.titleize : manifest.name
   end
 
+  def modified_at
+    # FIXME: ideally, we look at all the versioned files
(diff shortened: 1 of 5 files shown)


ericfranz:
Not doing right now.

Issue #202, opened 2017-07-18
Crash when app sharing disabled by I have sharing directory


nickjer:
My `.env.local`:

```bash
# .env.local

...

# comment out the below option
#OOD_APP_SHARING=true
```

And confirm I have a sharing directory:

```bash
$ ll -d /var/www/ood/apps/usr/jnicklas/
drwxr-xr-x 2 root root 4096 Sep 23  2016 /var/www/ood/apps/usr/jnicklas/
```

Load the Dashboard:

![screenshot from 2017-07-18 11-55-19](https://user-images.githubusercontent.com/4260509/28326909-0104ee34-6bb0-11e7-804e-76b9d6fd463f.png)

ericfranz:
So thats the Developer menu. That menu should not display if OOD_APP_SHARING is disabled.

ericfranz:
https://github.com/OSC/ood-dashboard/blob/e6d343ecd7895dac8ab0b9356f89d44c9589c421/app/apps/nav_config.rb#L7

The reason you experience this is that https://github.com/OSC/ood-dashboard/blob/e6d343ecd7895dac8ab0b9356f89d44c9589c421/config/initializers/ood.rb.osc.awesim#L14 is incomplete:

```ruby
# don't show develop dropdown unless you are setup for app sharing
NavConfig.show_develop_dropdown = UsrRouter.base_path.directory?
```

should change to

```ruby
NavConfig.show_develop_dropdown = ENV['OOD_APP_SHARING'].present? && UsrRouter.base_path.directory?
```

But the only portal this affects is AweSim. And AweSim's .env.local has OOD_APP_SHARING present.

nickjer:
You are right, it will only affect AweSim at the moment because `NavConfig.show_develop_dropdown` is only overridden in the AweSim initializer.

Another solution may be creating a custom getter in `NavConfig` such as:

```ruby
class NavConfig
  class << self
    attr_accessor :categories

    def show_develop_dropdown
      ENV['OOD_APP_SHARING'].present? && @show_develop_dropdown
    end

    attr_writer :show_develop_dropdown
  end
  self.categories = ["Files", "Jobs", "Clusters", "Interactive Apps"]
  self.show_develop_dropdown = true
end
```

ericfranz:
fixing #212 should fix this defect too, it just needs to be tested to determine that it fixes it

basilgohar:
[...]

PR #226, opened 2017-08-12
Separate app sharing and development env

### LINKED ISSUE #212: Separate enabling app sharing from enabling app development
Use OOD_APP_DEVELOPMENT separate from OOD_APP_SHARING for app development dropdown to appear

basilgohar:
Fixes #212 and #202

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -15,7 +15,7 @@ Rails.application.routes.draw do
   get "apps/icon/:name(/:type(/:owner))" => "apps#icon", as: "app_icon", defaults: { type: "sys" }
 
   #FIXME: undo when ready to deploy app sharing to production, remove?
-  if ENV['OOD_APP_SHARING'].present?
+  if ENV['OOD_APP_DEVELOPMENT'].present?
     # TODO:
     # is there a cleaner approach to this? an app should be a resource
     get "apps(/index(/:type(/:owner)))" => "apps#index", as: "apps", defaults: { type: "usr" }
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -15,6 +15,11 @@ Rails.application.routes.draw do
   get "apps/icon/:name(/:type(/:owner))" => "apps#icon", as: "app_icon", defaults: { type: "sys" }
 
   if ENV['OOD_APP_SHARING'].present?
+    # TODO:
+    # is there a cleaner approach to this? an app should be a resource
+    get "apps(/index(/:type(/:owner)))" => "apps#index", as: "apps", defaults: { type: "usr" }
+    get "apps/restart" => "apps#restart"
+
     root "apps#index", defaults: { type: "usr" }
   else
     root "dashboard#index"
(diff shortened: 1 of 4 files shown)


config/routes.rb
```diff
@@ -15,7 +15,7 @@
   get "apps/icon/:name(/:type(/:owner))" => "apps#icon", as: "app_icon", defaults: { type: "sys" }
 
   #FIXME: undo when ready to deploy app sharing to production, remove?
-  if ENV['OOD_APP_SHARING'].present?
+  if ENV['OOD_APP_DEVELOPMENT'].present?
```
ericfranz: 1. this should probably be NavConfig.show_develop_dropdown instead
2. the routes need to be split up into two groups, one for app sharing, and one for app development, and the separate flags should enable both; the way this PR currently is, OOD_APP_SHARING only does one thing: show a link in the develop dropdown. (this is wrong)

ericfranz:
I know that `NavConfig.show_develop_dropdown` might seem weird because we probably should have something like `AppConfig.app_development_enabled?` and `AppConfig.app_sharing_enabled?`. We'll visit that with the runtime only config etc.

basilgohar:
See c0d59a4 for the fix for the dashboard.  For app sharing, the shared apps are listed in the dashboard.  I believe all three cases are now possible with the above.  OOD_APP_SHARING present/not present, OOD_APP_DEVELOPMENT present/not present, and all combinations of such.  We can discuss tomorrow if you like.

config/routes.rb
```diff
@@ -14,15 +14,19 @@
   get "apps/show/:name(/:type(/:owner))" => "apps#show", as: "app", defaults: { type: "sys" }
   get "apps/icon/:name(/:type(/:owner))" => "apps#icon", as: "app_icon", defaults: { type: "sys" }
 
-  #FIXME: undo when ready to deploy app sharing to production, remove?
   if ENV['OOD_APP_SHARING'].present?
+    root "apps#index", defaults: { type: "usr" }
```
ericfranz: If `OOD_APP_SHARING` is present and `OOD_APP_DEVELOPMENT` is not present, then we are setting the root to "apps#index" but "apps#index" route will not exist because that is defined in the `if ENV['OOD_APP_DEVELOPMENT'].present?` block.

Since we can still enable `OOD_APP_DEVELOPMENT` now without `OOD_APP_SHARING`, and since in AweSim we already have `OOD_APP_DEVELOPMENT=1`, for the sake of time, I can still pull this in. Then I'll open a separate issue to revisit configuration of app development and app sharing.

config/routes.rb
```diff
@@ -44,7 +48,6 @@
       end
     end
   else
```
ericfranz: remove the unnecessary `else` here

config/routes.rb
ericfranz: Note this isn't an issue right now because we haven't advertised app sharing. But we did discuss "fixing" routes.rb so instead of `if ENV['OOD_APP_DEVELOPMENT'].present?` we would switch to using `if NavConfig.show_develop_dropdown` which would be replaced ultimately with a better feature flag like `AppConfig.developer_mode_enabled`. If and when we did that, the resulting code would crash the AweSim dashboard.

PR #229, opened 2017-08-15
Added a helpful `#to_bool` refinement

nickjer:
Added a helpful `#to_bool` refinement to Ruby objects that iHPC app developers can take advantage of. Follows the same code as:

https://github.com/rails/rails/blob/fbeebded22f53337df339285164352f298639c63/activemodel/lib/active_model/type/boolean.rb

Some examples:

```rb
"false".to_bool
#=> false

"0".to_bool
#=> false

"off".to_bool
#=> false

"".to_bool
#=> nil
```


ericfranz:
Should you also refine context?

```
TemplateBinding = Struct.new(:session, :context) 
```

Otherwise you open the door for problems:

```
session.var.to_bool # => false
context.var.to_bool # => CRASH
```

nickjer:
It works the same way the following works...

```rb
class Test
  using SymbolizeKeys

  def test
    test = { "a" => "A" }
    test.symbolize_keys
  end
end
```

Since `context` is a local variable inside the `Session` object it now has access to the `#to_bool` method.

nickjer:
In the end it is probably best for the developer to set strings on their form elements and do simple comparisons:

```yml
# form.yml
attributes:
  is_needed: "1"
```

```erb
# script.sh

<%- if context.is_needed == "1" -%>
do stuff
<%- end -%>
```

as this is very simple for most non-Ruby developers to understand.

PR #244, opened 2017-09-12
Replace product headers with breadcrumbs

ericfranz:


ericfranz:
Might be worth renaming the `products/header` partial to `products/breadcrumbs`. Otherwise I'm happy with it.

nickjer:
I agree with renaming it to `products/breadcrumbs` as we have a `<header>` in the layout and it makes me think we are replacing that.

Issue #236, opened 2017-08-29
Add AppConfig object and app development and app sharing flags


ericfranz:
In the dashboard we need a central place to store dashboard app specific configuration. This is domain configuration (i.e. whether app development is enabled or not, whether certain special URLs are set, etc.).

Currently these environment variables are just used throughout the app instead of being used to set configuration options in a central place. Similar to how we do in gems, we need a namespace for domain specific configuration, preferably separate from the Rails.application.configuration object.

While there are plenty of "examples" and "tutorials" on adding more config options to Rails.application.configuration i.e. Rails.application.configuration.x.xyz this is not good for two reasons:

1. the configuration options that can be set are not defined anywhere (a separate config class would define these as methods on the class)
2. custom Rails.application.configuration options are typically set on initializers, meaning we would have to ensure that our creation of these configuration options (with sensible defaults) occurred prior to other initializers that want to override these, or would have to occur after to ensure that we don't override custom values

Given this it makes the most sense we just have our own config object. However, if/when we do this we should be careful. Doing something like we do in NavConfig, where we set default values during class initialization, instead of returning default values lazily when methods are called, can result in these default values being re-set when Rails reloads the class in development mode. In other words: https://github.com/OSC/ood-dashboard/blob/ca4864b044f24b313b05f38a1fa3a6bb3249584c/app/apps/nav_config.rb#L6-L7 is re run every time code in this file is modified, but any initializers that set these values would not be rerun.

ericfranz:
Also we might want to think of a good name so we don't confuse this with Rails.application.configuration...

I'm not sure if AppConfig is sufficient.

nickjer:
I am fine with an app-specific configuration object, especially since I would like to move away from `.env.local` (using environment variables) and instead to a global app configuration file, e.g., `/etc/ood/config/apps/dashboard.yml.erb` (or something similar).

One possibility is...

The object can be defined in `config/configuration.rb` (this won't be reloaded by Rails). And we can add it inside of `config/environment.rb` as:

```ruby
# Load the Rails application.
require File.expand_path('../application', __FILE__)

# Load the application configuration.
require File.expand_path('../configuration', __FILE__)

# Initialize the Rails application.
Rails.application.initialize!
```

Then you can take advantage of it within the initializers.

ericfranz:
@nickjer good suggestion. PR created.

Issue #235, opened 2017-08-29
Crash when OOD_APP_DEVELOPMENT not set and NavConfig.show_develop_dropdown set to true


ericfranz:
Steps:

1. Make sure OOD_APP_DEVELOPMENT is not enabled
2. In initializer, do NavConfig.show_develop_dropdown
3. Navigating to dashboard causes crash

Problem: routes won't exist unless OOD_APP_DEVELOPMENT is specified

Solution?: verify routes.rb is run AFTER initializers; if this is the case, initializers can optionally configure the AppConfig (or NavConfig) and then the routes can use the feature flag in this config to enable the routes for app development

ericfranz:
Related to: https://github.com/OSC/ood-dashboard/issues/236

ericfranz:
Fixed by https://github.com/OSC/ood-dashboard/pull/247

PR #247, opened 2017-09-12
New config obj

ericfranz:
Fixes https://github.com/OSC/ood-dashboard/issues/236 and https://github.com/OSC/ood-dashboard/issues/235

Provides a new Configuration object that we can eventually:

1. move all domain specific configuration to
2. set sensible defaults from env vars
3. load a YAML config from /etc/config to set these values using a single YAML file instead of separate ENV vars

For now, only added app_development_enabled? and app_sharing_enabled?

### DIFF
--- a/config/configuration.rb
+++ b/config/configuration.rb
@@ -1,0 +1,17 @@
+# dashboard app specific configuration
+class Configuration
+  class << self
+    attr_writer :app_development_enabled
+    attr_writer :app_sharing_enabled
+
+    def app_development_enabled?
+      return @app_development_enabled if defined? @app_development_enabled
+      @app_development_enabled = ENV['OOD_APP_DEVELOPMENT'].present?
+    end
+
+    def app_sharing_enabled?
+      return @app_sharing_enabled if defined? @app_sharing_enabled
+      @app_sharing_enabled = ENV['OOD_APP_SHARING'].present?
+    end
+  end
+end
--- a/app/apps/nav_config.rb
+++ b/app/apps/nav_config.rb
@@ -1,8 +1,6 @@
 class NavConfig
   class << self
     attr_accessor :categories
-    attr_accessor :show_develop_dropdown
   end
   self.categories = ["Files", "Jobs", "Clusters", "Interactive Apps"]
-  self.show_develop_dropdown = ENV['OOD_APP_DEVELOPMENT'].present?
 end
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -29,7 +29,7 @@
         </ul>
         <div class="navbar-right">
           <ul class="nav navbar-nav">
-            <%= render partial: 'layouts/nav/develop_dropdown' if NavConfig.show_develop_dropdown %>
+            <%= render partial: 'layouts/nav/develop_dropdown' if Configuration.app_development_enabled? %>
             <%= render partial: "layouts/nav/help_dropdown" %>
           </ul>
           <p class="navbar-text" data-toggle="popover" data-content="Logged in as <%= @user.name %>" data-placement="bottom" >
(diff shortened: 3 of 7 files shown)


config/configuration.rb
```diff
@@ -0,0 +1,17 @@
+# dashboard app specific configuration
+class Configuration
+  class << self
+    attr_writer :app_development_enabled
+    attr_writer :app_sharing_enabled
+
+    def app_development_enabled?
+      return @app_development_enabled if defined? @app_development_enabled
+      @app_development_enabled = ENV['OOD_APP_DEVELOPMENT'].present?
```
nickjer: Another option is setting this in an initializer to maintain backwards compatibility for now:

```ruby
# config/initializers/read_env_vars.rb

Configuration.app_development_enabled = ENV["OOD_APP_DEVELOPMENT"].present?
Configuration.app_sharing_enabled = ENV["OOD_APP_SHARING"].present?
```

This will make it easier to remove in the future when it is deprecated.

config/configuration.rb
```diff
@@ -0,0 +1,17 @@
+# dashboard app specific configuration
+class Configuration
+  class << self
+    attr_writer :app_development_enabled
```
nickjer: This suggestion is more optional...

Since the following method `Configuration.app_development_enabled?` is a conditional with a question mark and not a typical getter `Configuration.app_development_enabled`, maybe making the following methods: `Configuration.enable_app_development` and `Configuration.disable_app_development` might be more prudent than the setter `Configuration.app_development_enabled=`.

config/configuration.rb
nickjer: So the above method would just be...

```ruby
def app_development_enabled?
  @app_development_enabled
end
```

config/configuration.rb
nickjer: This would allow us to force `@app_development_enabled` to always be a boolean.

config/configuration.rb
nickjer: Although might want to rename that initializer to `01_read_env_vars.rb` so it is loaded earlier than all the other initializers.

Or patch it directly into `config/environment.rb` so it loads before the initializers.

config/configuration.rb
ericfranz: Will you clarify how this is "optimal" and "prudent"?

nickjer: Guarantees `Configuration.app_development_enabled?` always returns a boolean, e.g.:

```ruby
Configuration.app_development_enabled = "something"
Configuration.app_development_enabled? #=> "something"
```

also having a setter in my opinion makes me think I can call the getter...

```ruby
Configuration.app_development_enabled = true
Configuration.app_development_enabled #=> RAISES ERROR
```

especially if we add more configuration options in the future that have both setters and getters...

```ruby
Configuration.motd_path = "/etc/motd"
Configuration.motd_path #=> "/etc/motd"
```

ericfranz: > Guarantees Configuration.app_development_enabled? always returns a boolean

In this case I don't think its a problem if it returns a boolean or an object since every object in Ruby has a truthy or falsy value.

I was asking for clarification because I thought your primary motivation for suggesting this change was:

> also having a setter in my opinion makes me think I can call the getter

This is just a preference and not a matter of "prudence" or "optimism".

config/configuration.rb
ericfranz: > This will make it easier to remove in the future when it is deprecated.

That assumes we will remove it.

config/configuration.rb
nickjer: > not a matter of "prudence"

Hah, not sure debating the meaning of "prudent" is necessary here. I just wanted to point out that this may or may not lead to confusion in the future:

```ruby
Configuration.motd_path = "/etc/motd"
Configuration.motd_path #=> "/etc/motd"

Configuration.app_development_enabled = true
Configuration.app_development_enabled #=> RAISES ERROR
```

But as I initially said, it is just a suggestion. So you can just say it is not necessary for now and ignore it.

config/configuration.rb
nickjer: I am fine with it as is. Didn't know if we slowly wanted to deprecate `dotenv` in favor of only a configuration file as it would simplify code and documentation.

config/configuration.rb
ericfranz: Oops I misread your comment:

> This suggestion is more optional...

Thought you said:

> This suggestion is more optimal...

PR #253, opened 2017-09-20
Two Factor Authentication URL and Update Font Awesome

ericfranz:


nickjer:
Not sure what the issue was, but maybe Font Awesome's fixed width style would help:

http://fontawesome.io/examples/#fixed-width

app/assets/stylesheets/application.css.scss
```diff
@@ -59,6 +59,11 @@ pre.motd-monospaced {
   white-space: pre-wrap;
 }
 
+i.fa {
```
nickjer: Maybe this style should be labeled under a unique class, like `fa-center-14px`? As this will affect all Font Awesome icons in the Dashboard.

PR #265, opened 2017-10-12
Use a login shell on LSF for Interactive Apps

### LINKED ISSUE #262: Fix shell setup for LSF
For LSF, set the login shell when submitting jobs:

https://github.com/OSC/ood-dashboard/blob/a65d101b1e339d3644a764935dedf301faf9be77/app/models/batch_connect/session.rb#L230-L239

When doing this, the adapter itself should set -env "all" if using LSF 9+. Also, we should verify that LSF 9+, when using -L "/bin/bash" and -env "all" at the same time that environment variables get passed through.

Another alternative would be to not have these adapter specific options in https://github.com/OSC/ood-dashboard/blob/a65d101b1e339d3644a764935dedf301faf9be77/app/models/batch_connect/session.rb#L230-L239 and instead have a "set login shell" option on the OodCore::Job::Script object. Slurm could ignore it, claiming "it sets the shell from shebang of the script". But LSF and Torque and PBS Pro adapters could add arguments appropriately. This might be the best...

nickjer:
Fixes #262

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- Load a proper login shell for Interactive Apps when using an LSF resource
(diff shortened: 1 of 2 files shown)


ericfranz:
- [ ] I need to test this change at TSC, in particular, what happens in this case: `bsub -L /bin/bash -L /bin/bash`

PR #268, opened 2017-10-13
Enable dev dashboard if sandbox exists

### LINKED ISSUE #264: Enable developer mode by default with the existence of the dev directory
So we would change https://github.com/OSC/ood-dashboard/blob/a65d101b1e339d3644a764935dedf301faf9be77/config/configuration.rb#L9

to be something like:

```ruby
@app_development_enabled = ENV['OOD_APP_DEVELOPMENT'].present? || Pathname.new(Dir.home).join(ENV['OOD_PORTAL'], 'dev')
```

This would make our tutorials easier. For admins, enabling app development is just creating a directory in their home directory.

Of course, this would need to be well-documented.

Also we need a method that produces this so we don't duplicate this everywhere:

```ruby
Pathname.new(Dir.home).join(ENV['OOD_PORTAL'], 'dev')
```

nickjer:
Fixes #264

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Updated `ood_core` library which includes bug fixes and new Batch Connect
   helper functions for scripts.
   [#263](https://github.com/OSC/ood-dashboard/pull/263)
+- The developer dashboard is enabled by default if the sandbox directory exists
+  for the user. [#264](https://github.com/OSC/ood-dashboard/issues/264)
 
 ## [1.17.0] - 2017-09-27
 ### Added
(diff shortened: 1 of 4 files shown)


nickjer:
If we had the routes always enabled, then we could make this more dynamic. Meaning a user could create the directory and the Dashboard would update on the fly. This is possible with:

```ruby
    def app_development_enabled?
      return @app_development_enabled if defined? @app_development_enabled
      ENV['OOD_APP_DEVELOPMENT'].present? || DevRouter.base_path.exist?
    end
```

and with the routes enabled.

Otherwise the user needs to create the directory and then restart their PUN.

ericfranz:
For this, the products controller should get a before action that checks whether or not app development is enabled.

Something like:

```ruby
before_action :authorize_developer

def authorize_developer
  unless Configuration.app_development_enabled?
    redirect_to root_path, :alert => "App development mode is not enabled"
  end
end

nickjer:
Hmmm... I found this:

https://robots.thoughtbot.com/constrain-yourself

This will reproduce the previous behavior, but be more dynamic.

PR #279, opened 2017-10-26
Runtime only config

### LINKED ISSUE #284: Runtime only config in /etc
Add support for runtime only config in /etc

ericfranz:
Fixes #284


ericfranz:
@nickjer this and https://github.com/OSC/ood-myjobs/pull/244 are related. Took some ideas from your PR prototypes.

I think the scope of loading .env files from /etc is less than converting all the configuration options in the Dashboard and My Jobs apps to load from a YAML, so this and myjobs changes are related incremental changes. IMHO we will eventually move to most or completely YAML + shared env loaded in profile.

Notable changes:

1. RAILS_RELATIVE_URL_ROOT is now set to a default value in bin/setup, not in .env.production
2. dotenv is loaded in config/boot.rb which loads config/config_root.rb which has no dependency on Rails and can also be required by bin/setup-production in myjobs
3. config/configuration.rb includes ConfigRoot module so it gets the config_root method to use elsewhere
4. guard clauses can now be removed from config/configuration.rb  and it can use Rails, Rails.application, and domain models like DevRouter; this will make it easier in the future to convert YAML config to configuration option objects (instead of just strings) when necessary (i.e. Help links)
5. OOD_APP_CONFIG is used for the env var instead of OOD_CONFIG
6. any files in OOD_APP_CONFIG that override the local files mirror their relative position i.e. in the root of OOD_APP_CONFIG can be .env and .env.production, initializers go in OOD_APP_CONFIG/config/initializers/, views go in OOD_APP_CONFIG/app/views/ etc. This means we could probably add OOD_APP_CONFIG/lib/... to the load path for the app in the future though I didn't test that yet.

ericfranz:
Forgot the important parts of "runtime only config":

1. `ood_appkit/bootstrap-overrides` is replaced with the bootstrap defaults in `app/assets/stylesheets/application.css.scss`
2.  app/views/layouts/application.html.erb renders a partial for inline styles to override navbar colors, based on `Configuration.custom_brand_bg_color` and `Configuration.custom_brand_link_active_bg_color` (Configuration.custom_brand_link_active_bg_color could probably be calculated from Configuration.custom_brand_bg_color using SASS libraries to darken by 20% but parsing rgb or # color strings using regex require slightly more time than I had at the moment.)
3. Configuration.custom_brand_bg_color is one color, regardless of the type of navbar (so it will get BOOTSTRAP_NAVBAR_DEFAULT_BG or BOOTSTRAP_NAVBAR_INVERSE_BG or OOD_BRAND_BG_COLOR value)

ericfranz:
Note it seems that whether I do

    config.paths["config/initializers"] << "/users/PZS0562/efranz/awesim/config/dashboard/config/initializers"

or

    config.paths["config/initializers"].unshift "/users/PZS0562/efranz/awesim/config/dashboard/config/initializers"

It matters not, initializers in /etc run first, initializers in the local app path run second, so you can't override values. I thought it was based on the naming order, but that doesn't seem to be the case, though I did not do much testing with this.

config/configuration.rb
```diff
@@ -19,5 +24,21 @@ def app_sharing_enabled?
       @app_sharing_enabled = ENV['OOD_APP_SHARING'].present?
     end
     alias_method :app_sharing_enabled, :app_sharing_enabled?
+
+    def custom_brand_bg_color
+      return @custom_brand_bg_color if defined? @custom_brand_bg_color
```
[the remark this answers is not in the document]
nickjer: Why is this called `custom_brand_bg_color` instead of just `brand_bg_color`?

config/configuration.rb
```diff
@@ -19,5 +24,21 @@ def app_sharing_enabled?
...
+      @custom_brand_bg_color = ENV.values_at('OOD_BRAND_BG_COLOR', 'BOOTSTRAP_NAVBAR_DEFAULT_BG', 'BOOTSTRAP_NAVBAR_INVERSE_BG').compact.first
+    end
+
+    def custom_brand_link_active_bg_color
+      return @custom_brand_link_active_bg_color if defined? @custom_brand_link_active_bg_color
```
nickjer: Same as above, not sure we need to cache this. And possibly renaming this to `brand_link_active_bg_color`.

nickjer:
I'd hold off on merging this in until you get myjobs working in a similar fashion.

config/config_root.rb
```diff
@@ -0,0 +1,63 @@
...
+  # the .env and .env.production or .env.development files.
+  #
+  # Doing this in two separate loads means OOD_APP_CONFIG can be specified in
+  # the .env.local file, which will specify where to look for the /etc dotenv
+  # files. The default for OOD_APP_CONFIG is /etc/ood/config/apps/myjobs and
+  # both .env and .env.production will be searched for there.
+  def load_dotenv_files
+    Dir.chdir app_root do
+      # .env.local first, so it can override OOD_APP_CONFIG
+      Dotenv.load(*dotenv_local_files) unless dotenv_local_files.empty?
```
nickjer: Due to the strange complexity this introduces just to get at `OOD_APP_CONFIG`, why don't we just introduce a new env file such as...

```ruby
Dotenv.load([app_root.join(".env.config")])
```

or

```ruby
Dotenv.load([app_root.join(".env.all.local")])
```

or 

```ruby
Dotenv.load([app_root.join(".env.super.local")])
```

This way it isn't a *possible* breaking change (for those super rare edge cases) like:

```bash
# .env
A=${B}
```

```bash
# .env.local
B="should be local to this file only"
```

It should return:

```ruby
puts ENV["A"] #=> ""
```

but instead it returns:

```ruby
puts ENV["A"] #=> "should be local to this file only"
```

nickjer: This will really only ever be used by Dashboard developers, so I don't think it will be a problem.

nickjer:
Is `ConfigRoot` necessary? Would it be possible to get away with just `Configuration` here (with the `ConfigRoot` methods incorporated into it)?

You can make `Configuration` a standalone object similar to what I do in the JobComposer example PR I wrote:

https://github.com/OSC/ood-myjobs/pull/243

That way it can be included in the `bin/setup-production` and Rails. At the bottom of the `config/configuration.rb` you can do something like...

```ruby
require "active_support/core_ext/hash/compact"

# config/configuration.rb
class Configuration
  # ...
end

Configuration.load_dotenv_files
```

ericfranz:
>Why is the file name _styles.html.erb as opposed to _styles.css.erb (possibly you should just include the <style></style> tags in the partial and keep the html name.

Rails complained about the `.css` extension so I changed it to html. :-)

This is the ramification of this change:

```diff
   def dotenv_files
     [
       config_root.join(".env.#{rails_env}"),
-      config_root.join(".env"),
       app_root.join(".env.#{rails_env}"),
+      config_root.join(".env"),
       app_root.join(".env")
     ].compact
   end
```

If I set this in `config_root.join(".env")`:

```
OOD_DATAROOT=$HOME/awesim/data/usr/efranz/myjobs
DATABASE_PATH=$OOD_DATAROOT/production.sqlite3
```

it will be ignored and `app_root.join(".env.production")` will be used instead. So anything set in `app_root.join(".env.production")` I would override in `config_root.join(".env.production")`. Which is okay, because if we don't make the above change in the load order, then the custom OOD_DATAROOT and DATABASE_PATH need to be set in `config_root.join(".env.production")` anyways, otherwise, appearing in `config_root.join(".env")` it would take precendence over `app_root.join(".env.development")`.

So in both cases setting production only values for env vars in in /etc/ will need to be done in `config_root.join(".env.production")` and env vars that are not RAILS_ENV dependent would be in `config_root.join(".env")`.

config/configuration.rb
ericfranz: Done: 72c57d3

config/configuration.rb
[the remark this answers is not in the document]
ericfranz: Done: 72c57d3

ericfranz:
Talked offline. The solution would be to **remove** .env.production, and have whatever is specified there used as sensible defaults in the app.

config/boot.rb
```diff
@@ -1,3 +1,8 @@
+# Set up gems listed in the Gemfile.
 ENV['BUNDLE_GEMFILE'] ||= File.expand_path('../../Gemfile', __FILE__)
 
 require 'bundler/setup' # Set up gems listed in the Gemfile.
+
+# load dotenv files before "before_configuration" callback
+require File.expand_path('../config_root', __FILE__)
+ConfigRoot.load_dotenv_files
```
nickjer: Maybe I forgot, but what is the reason for including these two lines in `boot.rb` as opposed to the head of the `configuration.rb` file?

nickjer:
> Talked offline. The solution would be to **remove** `.env.production`, and have whatever is specified there used as sensible defaults in the app.

@ericfranz: I am second guessing myself on that now, after some thinking I am leaning towards the following solution:

```ruby
module ConfigRoot
  ...

  def dotenv_files
    [
      config_root.join("env"),
      app_root.join(".env.#{rails_env}"),
      app_root.join(".env")
    ].compact
  end
end
```

So the sys admin would only ever need to modify a **single** `/etc/.../env` file similar to what he/she is used to previously, where they only ever modified a single `.env.local`. Basically `/etc/.../env` now behaves like a pseudo `/path/to/app/.env.local`.

Upgrading becomes `mv /path/to/app/.env.local /etc/.../env` (they have to delete the old `.env.local`) and it will work (since the order of the files is maintained).

Installation becomes do nothing because our previous Dashboard app accommodated no `/path/to/app/.env.local`.

Developers can override anything set in `/etc/.../env` with their own custom `/path/to/app/.env.local` which is gitignored.

config/boot.rb
nickjer: So you'd have something like...

```ruby
# config/configuration.rb
require File.expand_path('../config_root', __FILE__)
ConfigRoot.load_dotenv_files

# dashboard app specific configuration
class Configuration
  extend ConfigRoot

  ...
end

...
```

nickjer: It could be shortened further by calling `ConfigRoot.load_dotenv_files` at the end of `config/config_root.rb` instead of outside of it and after the `require`.

nickjer: So that would look like:

```ruby
# config/config_root.rb
require 'pathname'
require 'dotenv'

module ConfigRoot
  ...
end

ConfigRoot.load_dotenv_files
```

and

```ruby
# config/configuration.rb
require File.expand_path('../config_root', __FILE__)

# dashboard app specific configuration
class Configuration
  extend ConfigRoot

  ...
end

# support custom initializers and views in /etc
Rails.application.configure do |config|
  config.paths["config/initializers"] << Configuration.config_root.join("config", "initializers").to_s
  config.paths["app/views"].unshift Configuration.config_root.join("app", "views").to_s
end
```

If this were Job Composer you would then have a `bin/setup-production` with...

```ruby
#!/usr/bin/env ruby
ENV['RAILS_ENV'] = 'production'  # run this in production

require 'pathname'
require 'fileutils'
require_relative '../config/config_root.rb'

...
```

ericfranz:
In that case, I'll just keep:

```
      config_root.join("env.#{rails_env}"),
      config_root.join("env"),
      app_root.join(".env.#{rails_env}"),
      app_root.join(".env")
```

They both work like .env.local and env.production.local files.

We can tell everyone just to use `config_root.join("env"),` but if you are doing development its good to know about `config_root.join("env.#{rails_env}")` and anything that directly modifies the dataroot in there.

The YAML version would be to have a dev test prod and a default sections where default has all the shared config and prod has the dataroot, database, other relevant information (if there will be any in the future, but those are the only examples I know of now).

config/boot.rb
ericfranz: It is required in boot because we have to

1. require it before before_configuration callback is called
2. require before all of the Rails application code is initialized to increase parity between the loading this file when launching an app, versus running a rake task, versus requiring in `bin/setup-production` (https://github.com/OSC/ood-myjobs/blob/eaf1085344886b6655bf55fd29b8a095c6348ac8/bin/setup-production#L9 is what bin/rake does so I do that instead of `require_relative '../config/config_root'
` which I was doing before)

I'm warming to the idea of merging config_root and configuration, instead of keeping separate. We'd move the rails config.paths initialization back to the application.rb then. But we would still load it in boot.rb, not in environment.rb.

nickjer:
Sorry, I am still unsure of why we are adding `config_root.join("env.#{rails_env}")`. This seems to be a new feature as we currently don't use the corresponding file `.env.local.#{rails_env}` in any of the system OOD apps (correct me if I am wrong).

So far all of the system OOD apps are designed to only expect custom configuration in `.env.local` (since it is the only env file added to `.gitignore`) which we will be moving to `/etc/ood/.../env`.

I feel adding `config_root.join("env.#{rails_env}")` for now might be overkill as we have no use case for it. I think any future addition/feature to global configuration should go towards a `config.yml`.

We can have this discussion offline if you want.

config/boot.rb
nickjer: Ah, makes sense putting it in `boot.rb` then.

> I'm warming to the idea of merging config_root and configuration, instead of keeping separate. We'd move the rails config.paths initialization back to the application.rb then. But we would still load it in boot.rb, not in environment.rb.

I am for this.

ericfranz:
- [ ] TODO: verify when (or how) `Rails.env` is set => can we use this instead of a custom method?

config/application.rb
```diff
@@ -31,5 +31,8 @@ class Application < Rails::Application
 
     # Custom error pages
     config.exceptions_app = self.routes
+
+    config.paths["config/initializers"] << AppConfig.initializers_root.to_s if AppConfig.load_external_config?
+    config.paths["app/views"].unshift AppConfig.config_root.join("views").to_s if AppConfig.load_external_config?
```
nickjer: Maybe `AppConfig.config_root.join("views")` should be a method on `AppConfig` (e.g., `AppConfig.views_root`).

config/boot.rb
```diff
@@ -1,3 +1,14 @@
+# Set up gems listed in the Gemfile.
 ENV['BUNDLE_GEMFILE'] ||= File.expand_path('../../Gemfile', __FILE__)
 
 require 'bundler/setup' # Set up gems listed in the Gemfile.
+
+# load dotenv files before "before_configuration" callback
+require File.expand_path('../configuration', __FILE__)
+
+# global instance to access and use
+AppConfig = Configuration.new
```
nickjer: The `AppConfig` variable doesn't follow typical Ruby naming conventions:

https://github.com/bbatsov/ruby-style-guide#naming

They recommend camel case for `Class` or `Module` only.

bin/setup
```diff
@@ -7,6 +7,9 @@ include FileUtils
 ENV["RAILS_ENV"]               ||= ENV["PASSENGER_APP_ENV"] || "development"
 ENV["RAILS_RELATIVE_URL_ROOT"] ||= ENV["PASSENGER_BASE_URI"]
 
+# if relative url root not set, use default for this app
+ENV['RAILS_RELATIVE_URL_ROOT'] ||= "/pun/sys/dashboard" if ENV['RAILS_ENV'] == 'production'
+
```
nickjer: Consider removing all usages of `OOD_PORTAL` and `OOD_SITE` in this file as they are unnecessary now.

config/configuration.rb
```diff
@@ -1,23 +1,115 @@
...
+  # Load the dotenv local files first, then the /etc dotenv files and
+  # the .env and .env.production or .env.development files.
+  #
+  # Doing this in two separate loads means OOD_APP_CONFIG_ROOT can be specified in
+  # the .env.local file, which will specify where to look for the /etc dotenv
+  # files. The default for OOD_APP_CONFIG_ROOT is /etc/ood/config/apps/myjobs and
+  # both .env and .env.production will be searched for there.
+  def load_dotenv_files
+    # .env.local first, so it can override OOD_APP_CONFIG_ROOT
+    Dotenv.load(*dotenv_local_files) unless dotenv_local_files.empty?
```
nickjer: The `dotenv_local_files` array will never be empty so the `unless ...` is unnecessary.

config/configuration.rb
```diff
@@ -1,23 +1,115 @@
...
+  def config_root
+    Pathname.new(ENV["OOD_APP_CONFIG_ROOT"] || "/etc/ood/config/apps/dashboard")
+  end
+
+  def initializers_root
+    Pathname.new(ENV["OOD_APP_INITIALIZERS_ROOT"] || config_root.join("initializers"))
+  end
+
+  def load_external_config?
+    to_bool(ENV.fetch('OOD_LOAD_EXTERNAL_CONFIG', (rails_env == 'production')))
```
nickjer: More of a stylistic opinion, but this is also equivalent to:

```ruby
to_bool(ENV["OOD_LOAD_EXTERNAL_CONFIG"] || rails_env == "production)
```

So choose whichever you think is easiest to read.

config/configuration.rb
```diff
@@ -1,23 +1,115 @@
...
+    # .env.local first, so it can override OOD_APP_CONFIG_ROOT
+    Dotenv.load(*dotenv_local_files) unless dotenv_local_files.empty?
+
+    # load the rest of the dotenv files
+    Dotenv.load(*dotenv_files)
+  end
+
+  def app_development_enabled?
+    return @app_development_enabled if defined? @app_development_enabled
+    to_bool(ENV.fetch('OOD_APP_DEVELOPMENT', DevRouter.base_path.exist?))
```
nickjer: Same style opinion as above, this is equivalent to:

```ruby
to_bool(ENV["OOD_APP_DEVELOPMENT"] || DevRouter.base_path.exist?)
```

Whichever you prefer is fine.

config/configuration.rb
```diff
@@ -1,23 +1,115 @@
...
+  end
+  alias_method :app_development_enabled, :app_development_enabled?
+
+  def app_sharing_enabled?
+    return @app_sharing_enabled if defined? @app_sharing_enabled
+    @app_sharing_enabled = to_bool(ENV['OOD_APP_SHARING'])
```
nickjer: The assignment `@app_sharing_enabled = ...` isn't necessary here. See the above `app_development_enabled?` method.

config/configuration.rb
```diff
@@ -1,23 +1,115 @@
...
-      ENV['OOD_APP_DEVELOPMENT'].present? || DevRouter.base_path.exist?
-    end
-    alias_method :app_development_enabled, :app_development_enabled?
-
-    def app_sharing_enabled?
-      return @app_sharing_enabled if defined? @app_sharing_enabled
-      @app_sharing_enabled = ENV['OOD_APP_SHARING'].present?
-    end
-    alias_method :app_sharing_enabled, :app_sharing_enabled?
+  attr_writer :app_development_enabled
```
nickjer: This setter *may* not be necessary since we allow this to be set through environment variables (may require a little elbow grease to make env vars that call bash code to check if usr app path exists if a site wants that level of customization).

Also I worry that we allow configuration options to be mutated later in the code, but we use these configuration options when building the app. Maybe we make `Configuration` objects immutable?

config/configuration.rb
```diff
@@ -1,23 +1,115 @@
...
+  attr_writer :app_sharing_enabled
```
nickjer: See the above comment on mutability concerns.

config/configuration.rb
```diff
@@ -1,23 +1,115 @@
...
+    # copied from OodAppkit::Configuration#set_default_configuration
+    # then modified to ensure dataroot is never nil
+    #
+    # FIXME: note that this would be invalid if the dataroot where
+    # overridden in an initializer by modifying OodAppkit.dataroot
+    # Solution: in a test, add a custom initializer that changes this, then verify it has
+    # no effect or it affects both.
+    #
+    root = ENV['OOD_DATAROOT'] || ENV['RAILS_DATAROOT']
+    root ||= "~/#{ENV['OOD_PORTAL'] || 'ondemand'}/data/#{ENV['APP_TOKEN'] || 'sys/dashboard'}"
```
nickjer: What if we change this to:

```ruby
root ||= if rails_env == "production"
           "~/#{ENV['OOD_PORTAL'] || 'ondemand'}/data/#{ENV['APP_TOKEN'] || 'sys/dashboard'}"
         else
           app_root.join("data")
         end
```

That way we can remove `.env.development` and `.env.test`.

config/configuration.rb
```diff
@@ -1,23 +1,115 @@
...
+  private
+
+  # The environment
+  # @return [String] "development", "test", or "production"
+  def rails_env
+    ENV['RAILS_ENV'] || ENV['RACK_ENV'] || "development"
+  end
+
+  # The app's root directory
+  # @return [Pathname] path to configuration root
```
nickjer: Replace `configuration` with `app` in comment.

config/examples/README
```diff
@@ -0,0 +1,6 @@
+```
+mkdir -p /etc/ood/config/apps
```
nickjer: Add `sudo` maybe?

config/configuration.rb
nickjer: If you feel it is necessary, then you may need to add the assignment to the method `app_development_enabled?`.

config/examples/awesim/initializers/ood.rb
```diff
@@ -11,8 +11,8 @@
 end
 
 # don't show develop dropdown unless you are setup for app sharing
-Configuration.app_development_enabled = UsrRouter.base_path.directory?
-Configuration.app_sharing_facls_enabled = true
+AppConfig.app_development_enabled = UsrRouter.base_path.directory?
```
nickjer: This is a breaking change. Are there any OOD partners that use a custom initializer that makes calls to `Configuration`? If not, then this should be fine.

config/examples/README
ericfranz: 9ba768c

config/configuration.rb
ericfranz: cf6d1b8

config/configuration.rb
ericfranz: cf6d1b8

app/views/layouts/application.html.erb
```diff
@@ -10,6 +10,9 @@
   <%= yield :head %>
 
   <meta name="viewport" content="width=device-width, initial-scale=1">
+<style>
```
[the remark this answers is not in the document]
ericfranz: 329f2b3

config/application.rb
```diff
@@ -31,5 +31,8 @@ class Application < Rails::Application
 
     # Custom error pages
     config.exceptions_app = self.routes
+
+    config.paths["config/initializers"] << AppConfig.initializers_root.to_s if AppConfig.load_external_config?
```
nickjer: I like what you do in MyJobs with:

```ruby
if AppConfig.load_external_config?
  ...
end
```

bin/setup
ericfranz: 04c7482

config/application.rb
ericfranz: Or we get rid of initializers_root. I'm not sure what a separate configurable initializers root or app views root is useful for. In all my testing use cases, the only thing I needed to vary was the app config root itself, so instead of being in /etc/ it was in a directory in my home directory so I could do local testing between awesim and ondemand easily while in development mode.

config/boot.rb
ericfranz: I'd have to refresh my visual picture of the ruby object model, but I think that doing:

```
AppConfig = Configuration.new
```

is functionally equivalent to creating AppConfig class and then specifying all the config options as class methods in a `class << self do` block.



So this particular case is a little grey.  I introduced it in 6ac2d861093ef9b5e281e28fb40986895e34e5f0:

Its the first idea proposed in
https://8thlight.com/blog/josh-cheek/2012/10/20/implementing-and-testing-the-singleton-pattern-in-ruby.html

The idea is you want to use the Singleton pattern but you don't want the complexity of the Singleton pattern and you want the class you are making a singleton of easy to test.

ericfranz: In otherwords, here I'm defining a new class called AppConfig.

nickjer: We discussed this offline, I am fine with this since it makes testing easier.

config/application.rb
nickjer: I am fine with either decision you make, but should probably be consistent between both `initializers` and `views`.

Although some pros for instance methods:

- if we ever do change the location of those paths it would be easier to support backwards compatibility (or if we wanted to check for the existence of multiple locations)
- we can move the logic involving `rails_env` to the Configuration object itself as:

  ```ruby
  def initializers_root
    Pathname.new( ... ) if load_external_config?
  end
  ```

  and the `config/application.rb` can be updated with

  ```ruby
  config.paths["config/initializers"].push(AppConfig.initializers_root.to_s) if AppConfig.initializers_root
  ```

config/application.rb
nickjer: Note, if you go the route of moving Configuration logic to the Configuration object as discussed in a comment above I recently made, then this is moot.

config/examples/awesim/initializers/ood.rb
ericfranz: Now using `Configuration` for the singleton instance like before.

config/configuration.rb
ericfranz: Discussed offline. Thats a good concern. The original code provided a solution for this in that config_root and configuration where two separate objects, and only config_root was usable in bin/setup-production. I'll open a separate issue in myjobs on possibly adding some protection to ensure only immutable configuration options are used in the bin/setup-production script.

PR #281, opened 2017-10-27
[DO NOT MERGE] Simple example with configuration in global location

nickjer:
@ericfranz here is an example that maintains backwards compatibility but allows for a YAML (ERB rendered) configuration file located at:

```
/etc/ood/config/apps/dashboard/config.yml
```

and any initializers under:

```
/etc/ood/config/apps/dashboard/initializers/**/*.rb
```

I do not cache the configuration as it is typically read from local disk (so should be fast), and this allows for real-time updating of the Dashboard. Although it wouldn't be hard to add caching if you prefer.

> **NOTE:**
> You can override the location of the above files by adding the following line to your `.env.local`:
>
> ```bash
> OOD_CONFIG=/path/to/new/config_root
> ```
>
> and it will look for:
>
> ```
> /path/to/new/config_root/config.yml
> /path/to/new/config_root/initializers/**/*.rb
> ```

### DIFF
--- a/app/models/motd_file.rb
+++ b/app/models/motd_file.rb
@@ -7,7 +7,7 @@ class MotdFile
   # Initialize the Motd Controller object based on the current user.
   #
   # @param [String] path The path to the motd file as a URI
-  def initialize(path = ENV['MOTD_PATH'])
+  def initialize(path = Configuration.motd_path)
     @motd_path = path
     @content = load(path)
   end
@@ -34,7 +34,7 @@ class MotdFile
   # @return [Object, nil] an MotdFormatter object that responds to `:to_partial_path`
   #                       `nil` if a file does not exist at the path.
   def formatter
-    case ENV['MOTD_FORMAT']
+    case Configuration.motd_format
       when 'osc'
         @motd = MotdFormatterOsc.new(self)
       when 'markdown'
--- a/config/configuration.rb
+++ b/config/configuration.rb
@@ -19,5 +19,45 @@ class Configuration
       @app_sharing_enabled = ENV['OOD_APP_SHARING'].present?
     end
     alias_method :app_sharing_enabled, :app_sharing_enabled?
+
+    # The app's configuration root directory
+    # @return [Pathname] path to configuration root
+    def config_root
+      Pathname.new(ENV["OOD_CONFIG"] || "/etc/ood/config/apps/dashboard")
+    end
+
+    # A hash describing the configuration read in from the app's configuration
+    # file
+    # @return [Hash{Symbol=>Object}] configuration hash
+    def config
+      yaml = config_root.join("config.yml")
+      config = yaml.exist? ? YAML.load(ERB.new(yaml.read, nil, "-").result) : {}
+
+      if config.is_a?(Hash)
+        config.compact.deep_symbolize_keys
+      else
+        Rails.logger.error("ERROR: Configuration file '#{yaml}' needs to define a Hash")
+        {}
+      end
+    rescue Psych::SyntaxError => e
+      Rails.logger.error("ERROR: YAML syntax error occurred while parsing #{yaml} - #{e.message}")
+      {}
+    end
+
+    # The path to the motd file
+    # @return [String, nil] motd path
+    def motd_path
+      ENV["MOTD_PATH"] || config[:motd_path].try(:to_s)
+    end
+
+    # Format to display motd
+    # @return [String, nil] motd format
+    def motd_format
+      ENV["MOTD_FORMAT"] || config[:motd_format].try(:to_s)
+    end
   end
+end
+
+Rails.application.configure do |config|
+  config.paths["config/initializers"] << Configuration.config_root.join("initializers").to_s
 end


ericfranz:
This works well for many of the dashboard's configuration options, in particular, those that can be completely removed from the dotenv files and don't appear in .env and .env.production in any of the apps.

However, this is problematic for those configuration options that do appear in .env and .env.production. Things like OOD_SITE, OOD_PORTAL, OOD_DATAROOT, DATABASE_PATH, etc.

OOD_DATAROOT is particularly concerning:

1. For the dashboard and job composer, we have the case of changing the dataroot from `$HOME/ondemand/$APP_TOKEN` to `$HOME/awesim/$APP_TOKEN` for job composer and the dashboard this is done two different ways right now (compare dashboard and job composer). 
2. We have one external site who has changed the dataroot of the job composer to `/scratch/user/$USER` as that is where users run and store job data at their site.

In both of these cases, the YAML would not work. Also, in at least job composer, bin/setup-production script would need to be replaced or updated to properly load the correct environment required for setting the data root.

https://github.com/OSC/ood_appkit/pull/48 effectively inserts the /etc/ env file in between the .env.local and the other .env in the load order when they are loaded during the require dotenv/rails-now. With that we would still need to explicitly specify the /etc location for the core apps to ease app development and change the way the bin/setup-production files load the environment.

ericfranz:
Actually the only reason we do `require: 'dotenv/rails-now'` is because we are setting RAILS_RELATIVE_URL_ROOT in a dotenv file and we need that set when building assets. But we might not need that anymore with runtime config...

nickjer:
> Things like `OOD_SITE`, `OOD_PORTAL`, `OOD_DATAROOT`, `DATABASE_PATH`, etc.

- `OOD_SITE`: Is this used anywhere aside from copying the `.env.local.*` files, which won't be needed after these code changes take effect?
- `OOD_PORTAL`:
   This can be added to the `configuration.rb` as:

   ```ruby
   class Configuration
     # Id describing the OnDemand portal
     # @return [String, nil] portal id
     def portal
        ENV["OOD_PORTAL"] || config[:portal].try(:to_s) || "ondemand"
     end
   end
   ```

   Then replace all calls of `ENV["OOD_PORTAL"]` in the Dashboard code with `Configuration.portal`.
- `OOD_DATAROOT`:
   This can added to the `configuration.rb` as:

   ```ruby
   class Configuration
     # Path to the data root for this app
     # @return [String, nil] data root
     def dataroot
        ENV["OOD_DATAROOT"] || config[:dataroot].try(:to_s) || (ENV["APP_TOKEN"] ? "~/#{portal}/data/#{ENV["APP_TOKEN"]}" : nil)
     end
   end

   OodAppkit.configure do |config|
     config.dataroot = Configuration.dataroot
   end
   ```
- `DATABASE_PATH`: not used for Dashboard (will need to take a look at the Job Composer app to get a better understanding of how to adapt this)

> For the dashboard and job composer, we have the case of changing the dataroot from `$HOME/ondemand/$APP_TOKEN` to `$HOME/awesim/$APP_TOKEN`

This is now handled in Puppet by adding the appropriate global Dashboard configuration file for the respective web host. So for `web04` (AweSim production) we have:

```yaml
# /etc/ood/config/dashboard/config.yml
---
portal: "awesim"
```

For an OSC OnDemand web host we can ignore configuration as it will default to `ondemand` for the portal.

> We have one external site who has changed the dataroot of the job composer to `/scratch/user/$USER` as that is where users run and store job data at their site.

This can now be set in the global Dashboard configuration file as:

```yaml
# /etc/ood/config/dashboard/config.yml
---
dataroot: "/scratch/user/<%= ENV["USER"] %>"
```

> Also, in at least job composer, `bin/setup-production` script would need to be replaced or updated to properly load the correct environment required for setting the data root.

You are right, but I have yet to take a deep look at the Job Composer app. When I get the time I can "try" to make an example PR for it as well.

ericfranz:
1. The yaml.erb approach will not work well with Python and Node.js. For those, presumably though, we just load /etc/ood/env first then the other location
2. My point about the `OOD_PORTAL` being set was we have a bunch of .env and .env.production files in our Rails apps, that set values based on `OOD_PORTAL`. We would need to stop this approach completely if we moved to YAML.
3. We still have a problem with ood_appkit gem depending on environment variables such as OOD_CLUSTERS and APP_TOKEN and OOD_DASHBOARD_TITLE etc. that this would not solve. But I think right now for those, besides APP_TOKEN, we do not have a use case for changing. OOD_CLUSTERS we change often when developing, but "when developing" can just mean use a .env.local file

nickjer:
> The yaml.erb approach will not work well with Python and Node.js

You can shell out `erb /etc/ood/config/<app>/config.yml` and then feed that to the YAML renderer.

Also, you cannot use Bash substitution for Node.js, so this will fail:

```bash
OOD_DATAROOT="/path/to/${USER}"
```

> My point about the `OOD_PORTAL` being set was we have a bunch of `.env` and `.env.production` files in our Rails apps, that set values based on `OOD_PORTAL`

I am confused by this, is there an example? Any environment variable set in `.env.local` is not available to `.env` or `.env.production`.

> We still have a problem with `ood_appkit` gem depending on environment variables such as `OOD_CLUSTERS` and `APP_TOKEN` and `OOD_DASHBOARD_TITLE` etc.

- `OOD_CLUSTERS`: Do we have a use-case in production where a sys admin would put the cluster configs in any other location? All of our documentation points to `/etc/ood/config/clusters.d`. Although see below for discussion on `/etc/ood/profile`.
- `APP_TOKEN`: This will be unnecessary for `dataroot` with the above patch. I haven't looked at how it is being used in the logger quite yet.
- `OOD_DASHBOARD_TITLE`: This can be overridden with something like...

   ```ruby
   OodAppkit.configure do |config|
     config.dashboard = Urls::Dashboard.new(
       title:    Configuration.dashboard_title,
       base_url: Configuration.dashboard_base_url
     )
   end
```

Although as you mentioned, some of these can be set as global env vars under `/etc/ood/profile` (renamed from `/etc/ood/env` since it behaves more like the `profile` script in linux). This will be loaded before NGINX is loaded, so something like:

```bash
# /etc/ood/profile

export OOD_CLUSTERS="/etc/ood/config/clusters.d"
export OOD_DASHBOARD_TITLE="Dashboard"
```

You can't set `APP_TOKEN` there because it is app-dependent.

ericfranz:
Suggestions from this have been included in https://github.com/OSC/ood-dashboard/pull/279.

PR #294, opened 2017-12-15
Add All Apps Index Page

ericfranz:


app/apps/ood_app.rb
```diff
@@ -40,6 +42,29 @@ def url
...
+      else
+        ApplicationController.helpers.login_clusters.map { |c| Link.new("#{c.metadata.title} Shell Access", OodAppkit.shell.url(host: c.login.host), "terminal") }
+      end
+    elsif role == "batch_connect"
+      batch_connect.sub_app_list.select(&:valid?).map do |sub_app|
+        Link.new(sub_app.title, Rails.application.routes.url_helpers.new_batch_connect_session_context_path(token: sub_app.token))
+      end
+    else
+      # normal, use default icon
+      [Link.new(title, Rails.application.routes.url_helpers.app_path(name, type, owner))]
```
nickjer: This should probably be:

```ruby
[Link.new(title, url)]
```

otherwise it won't read the url in the manifest.yml file for apps other than batch connect apps.

app/apps/ood_app.rb
```diff
@@ -40,6 +42,29 @@ def url
...
+    if role == "files"
+      # assumes Home Directory is primary...
+      [Link.new("Home Directory", OodAppkit.files.url(path: Dir.home), 'home')] + OodFilesApp.new.favorite_paths.map do |path|
+        Link.new(path.to_s, OodAppkit.files.url(path: path), "folder")
+      end
+    elsif role == "shell"
+      if ApplicationController.helpers.login_clusters.count == 0
+        [Link.new("Shell Access", OodAppkit.shell.url,"terminal")]
+      else
+        ApplicationController.helpers.login_clusters.map { |c| Link.new("#{c.metadata.title} Shell Access", OodAppkit.shell.url(host: c.login.host), "terminal") }
```
nickjer: This should probably be:

```ruby
ApplicationController.helpers.login_clusters.map do |c|
  Link.new("#{c.metadata.title || c.id.to_s.titleize} Shell Access", OodAppkit.shell.url(host: c.login.host), "terminal")
end
```

as `Cluster#metadata` has attributes that could be `nil`.

app/views/apps/_app_list.html.erb
```diff
@@ -0,0 +1,10 @@
+<div class="panel panel-default">
+  <div class="panel-heading"><%= app_list.title %></div>
+  <div class="list-group">
+
+    <% OodAppGroup.groups_for(apps: app_list.apps, group_by: :subcategory).each_with_index do |g, index| %>
+      <%= content_tag "p", g.title, class: "list-group-item header" unless g.title.empty? %>
+      <%= render partial: "apps/app_list_item", collection: g.apps, as: :app, locals: { current_app_token: local_assigns[:current_app_token] }%>
```
nickjer: Do we need to prepend "apps/" to the partial path? We didn't do that below for this partial.

app/views/apps/_app_list_item.html.erb
```diff
@@ -0,0 +1,15 @@
+<% app.links.each do |link| %>
+  <%= link_to link.url.to_s,
+    class: "list-group-item",
+    data: {
```
nickjer: Make this `Hash` optional if `app.manifest.description` exists.

app/views/apps/_app_list_item.html.erb
```diff
@@ -0,0 +1,15 @@
...
+      toggle: "popover",
+      content: manifest_markdown(app.manifest.description),
+      html: true,
+      trigger: "hover",
+      title: app.title,
```
nickjer: Maybe this should be `link.title` and not `app.title`?

app/views/apps/index.html.erb
```diff
@@ -1,17 +1,21 @@
...
-    <%= render @motd if @motd %>
+    <div class="panel panel-default">
+      <div class="panel-heading">System Apps</div>
+      <div class="panel-body">
+        <%= render partial: "app_list", collection: @core_app_groups %>
+      </div>
+    </div>
   </div>
+
+  <%= render partial: "app_list", collection: @app_groups, layout: "col3_layout" %>
```
nickjer: Not a bug, but we may need to be aware that if `@app_groups` grows beyond 3 groups, then we will overflow the Bootstrap row. At that point we may need to explore a more dynamic grid system.

app/views/apps/_app_list.html.erb
nickjer: I believe you can remove the `locals: { ... }` on this line as it isn't used in the corresponding partial.

config/examples/osc/env
```diff
@@ -17,3 +17,6 @@ OOD_BRAND_LINK_ACTIVE_BG_COLOR='#990c23'
 
 OOD_DASHBOARD_DOCS_URL="https://www.osc.edu/ondemand"
 OOD_DASHBOARD_PASSWD_URL="https://my.osc.edu"
+
+# show all apps index page
+SHOW_ALL_APPS_LINK=1
```
nickjer: Should this also be added to `config/examples/awesim/env` since we are keeping AweSim and OSC OnDemand in sync now?

app/views/apps/index.html.erb
ericfranz: I was betting on right now that the Bootstrap grid will just put extra columns on a separate row i.e. like it does with thumbnails:

>If more than 12 columns are placed within a single row, each group of extra columns will, as one unit, wrap onto a new line.

But the result is that produces an undesirable affect when one column is much larger that the others. The new row of columns starts after the longest column completes. Using something like https://masonry.desandro.com/ would make the columns fit better, but then they are not reliably in the same location every time.

At this point, a smaller table view, with one app per row, and a small row, might be preferable, with a text box for filtering i.e. datatables.

Benefit of having a separate Apps page is we can play with this, even provide different views  to switch between, without risking a jarring change to most user's day to day experience or affecting Open OnDemand (since changes to this doesn't affect the nav bar).

ericfranz: Of course the problem with a table view is you don't see as many apps at a time.

config/examples/osc/env
ericfranz: Yeah. If we do this, we should probably add shared apps (or featured apps) section at the top of this page... or to the far right after Gateway Apps.

app/views/apps/index.html.erb
ericfranz: FWIW I think it can actually fit 4 groups. Here was a test I did earlier but didn't try this after the change I made. This is 1024x768:

screen 2017-12-12 at 3 52 17 pm

nickjer: Masonry is actually nice since we have some very long columns and short ones. Also, how does it handle 5 columns, if there are 4 `@app_groups` + the `@core_app_groups`?

nickjer:
@ericfranz you are also missing changes to the `CHANGELOG.md`

nickjer:
Nevermind, I see you added it to the `CHANGELOG.md` outside of this PR.

PR #298, opened 2017-12-18
Add global config for batch connect apps

### LINKED ISSUE #293: Support global config for `local` batch_connect sub-apps
Currently batch_connect apps can support sub-apps (or overrides for the main app) in a directory called `local/` underneath the root of the app itself:

https://github.com/OSC/ood-dashboard/blob/d6833b7e7f73c0f3eacc06e53cf5c9955bf4604a/app/models/batch_connect/app.rb#L63-L67

This should be expanded to support our goals of transforming all OOD system apps to support global configuration. In particular this is for the OOD system app [bc_desktop](https://github.com/OSC/bc_desktop/).

nickjer:
Fixes #293

### DIFF
--- a/config/configuration_singleton.rb
+++ b/config/configuration_singleton.rb
@@ -36,14 +36,10 @@ class ConfigurationSingleton
     to_bool(ENV['OOD_LOAD_EXTERNAL_CONFIG'] || (rails_env == 'production'))
   end
 
-  # The batch_connect apps' configuration root directory
+  # The root directory that holds configuration information for Batch Connect
+  # apps (typically each app will have a sub-directory underneath this)
   def bc_config_root
     Pathname.new(ENV["OOD_BC_APP_CONFIG_ROOT"] || "/etc/ood/config/apps")
-  end
-
-  # Whether we load the batch_connect apps' configuration from external root
-  def load_external_bc_config?
-    to_bool(ENV["OOD_LOAD_EXTERNAL_BC_CONFIG"])
   end
 
   # Load the dotenv local files first, then the /etc dotenv files and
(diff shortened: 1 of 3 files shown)


config/configuration_singleton.rb
```diff
@@ -36,6 +36,16 @@ def load_external_config?
...
 
+  # The batch_connect apps' configuration root directory
+  def bc_config_root
+    Pathname.new(ENV["OOD_BC_APP_CONFIG_ROOT"] || "/etc/ood/config/apps")
+  end
+
+  # Whether we load the batch_connect apps' configuration from external root
+  def load_external_bc_config?
+    to_bool(ENV["OOD_LOAD_EXTERNAL_BC_CONFIG"])
+  end
```
ericfranz: Before I add comments, can you further explain your thinking on the intention of `load_external_bc_config?` and `bc_config_root` and then the logic when actually loading?

For example, the implementation of

    def load_external_config?
       to_bool(ENV['OOD_LOAD_EXTERNAL_CONFIG'] || (rails_env == 'production'))
     end

has these goals:

1. default functionality is `load_external_config?` is true when `rails_env == 'production'`
2. if the OOD_LOAD_EXTERNAL_CONFIG env is set we use the bool value of that string, whether it is yes/no 1/0 true/false. This allows me in Rails development mode to enable loading external config, and in Rails production mode disable loading external config using a .env.local file. The former more than the latter is really useful for development.

It seems that `load_external_bc_config?` doesn't yet serve that purpose because it is always false unless `OOD_LOAD_EXTERNAL_BC_CONFIG` is set to something; and if `load_external_bc_config?` is false we are still loading external config for sys apps if the sys app directory exists:

    if Configuration.load_external_bc_config? || (router.type == :sys && global_sub_app_root.directory?)

nickjer: Ah, I forgot that when the env var is `nil` we fallback to default decision-making, I am calling the `to_bool` too early here. I will need to rethink this.

The simplest solution is that I remove the `#load_external_bc_config?` method and just use:

```ruby
if router.type == :sys && global_sub_app_root.directory?
```

PR #303, opened 2017-12-19
Added OOD branding

### LINKED ISSUE #292: Add OOD branding to default install
With all the provided OOD branding material:

https://osu.box.com/s/pddtz424et66ki4meu1wd493h8gjr9jl

We should apply it to a default install of the Dashboard app.

nickjer:
Fixes #292 

image


ericfranz:
So did a test based on https://philipwalton.github.io/solved-by-flexbox/demos/sticky-footer/ and https://css-tricks.com/almanac/properties/f/flex/ and https://stackoverflow.com/questions/28389667/using-flexbox-sticky-footer-with-bootstrap.

Add style to body:

```
<body class="vsc-initialized" style="
    display: flex;
    min-height: 100vh;
    flex-direction: column;
">
```

and the main container:

```
<div class="container" role="main" style="
    flex: 1;
">
```

And sticky footer results.

Though, may need all the CSS vendor prefixes for flex and flex-grow. Also, https://stackoverflow.com/questions/28389667/using-flexbox-sticky-footer-with-bootstrap mentions a problem with this approach and rendering at < 768px width which I didn't observer with Chrome. Also, a change like this would require multi-browser testing. But we do use flex with the Files Explorer so...

nickjer:
I tested this in Chrome, IE11, and Firefox on Windows 10. It seems to work in all 3 browsers no matter the window dimensions.

nickjer:
It does not work on my Android phone with Chrome. I actually have to scroll down to see the footer. Not sure why though.

nickjer:
See code comments here:

https://github.com/philipwalton/solved-by-flexbox/blob/master/assets/css/components/site.css

Although I didn't notice a bug in IE11.

PR #310, opened 2017-12-29
Add support for YAML announcements

### LINKED ISSUE #306: Add support for a more advanced notification system
After a deep-dive discussion the following requirements for a Dashboad notification system were distilled:

- allow notification to be displayed in different locations with respect to Dashboard logo on OnDemand root page
- allow it to be displayed on different pages (all Dashboard pages, e.g. iHPC apps, or just OnDemand/AweSim root pages)
- allow it to be displayed with different colors (danger, warning, info, success)
- provide conditional on whether to display notification or not
- allow its message to be dynamically generated

The first two requirements are related to each other.

nickjer:
Fixes #306 

Decided to keep code changes to a minimum so cut back on a couple of requirements. I left off ability to place announcement above/below main logo as well as ability to specify what pages announcement appears on. Although these can be added now or in the future.

It provides support for:

- specifying color of announcement with `:type`
- specifying whether announcement appears or not by setting `:msg` to a string or `nil`
- backwards compatible with `/etc/ood/config/announcement.md`, but can support `/etc/ood/config/announcement.yml` or a collection of markdown/yaml announcements in `/etc/ood/config/announcements.d/`

Included an example announcement for quota.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -19,7 +19,7 @@ class ApplicationController < ActionController::Base
   end
 
   def set_announcements
-    @announcements = Announcements.all
+    @announcements = Announcements.all(Configuration.announcement_path)
   rescue => e
     logger.warn "Error parsing announcements: #{e.message}"
     @announcements = []
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -3,7 +3,7 @@ class ApplicationController < ActionController::Base
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
 
-  before_filter :set_user, :set_nav_groups, :set_announcement
+  before_filter :set_user, :set_nav_groups, :set_announcements
 
   def set_user
     @user = User.new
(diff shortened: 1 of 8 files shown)


app/controllers/application_controller.rb
```diff
@@ -18,10 +18,10 @@ def sys_app_groups
     @sys_app_groups ||= OodAppGroup.groups_for(apps: SysRouter.apps)
   end
 
-  def set_announcement
-    path = Pathname.new(ENV["OOD_ANNOUNCEMENT_PATH"] || "/etc/ood/config/announcement.md")
-    @announcement = path.read if path.file?
+  def set_announcements
+    @announcements = Announcements.all
```
ericfranz: A consideration.

1. The test for `Announcements.all` mocks `Configuration.announcement_path` because it knows that the implementation of  `Announcements.all` uses `Configuration.announcement_path`
2.  `Announcements.all` contains guard clauses to handle the case that `Configuration.announcement_path` is nil 


If you ensure `Configuration.announcement_path` never returns nil, you can get rid of the other guard clauses and test mock:

```ruby
+      [
 +        "/etc/ood/config/announcement.md",
 +        "/etc/ood/config/announcement.yml",
 +        "/etc/ood/config/announcements.d"
 +      ].map {|p| Pathname.new(p)}.detect(&:exist?) || Pathname.new("/etc/ood/config/announcement.md")
```


Then the code could be simplified by:

1. dropping `Announcement.all` and renaming `Announcement.parse(path)` to  `Announcement.all(path)`
2. change this line from `@announcements = Announcements.all` to `@announcements = Announcements.all(Configuration.announcement_path)`
3. remove the old for `all` that mocks `Configuration.announcement_path` 

Of course, `Announcements.all(Configuration.announcement_path)` is not as pretty as  `Announcements.all` but it does make  `Announcements.all` functional.

I also notice that `Configuration.announcement_path` could be "" if we set that value. I don't think that is a likely edge case, and even if that was the case, `Pathname.new("").expand_path` returns the current directory, which probably won't have an announcement file (and in fact, that could be a useful trick when developing; just set OOD_ANNOUNCEMENT_PATH to ""; maybe that should even be the default value in development so we can just drop those files in place).

nickjer:
@ericfranz you might want to squash merge since I just bunched up all the fixes into a single commit.

nickjer:
Also it now accepts a single path or an array of paths, and generates announcements from them all.

PR #320, opened 2018-01-23
Ignore apps with period in dir name

### LINKED ISSUE #313: Hide apps that have a period in the name
It is common at least in Linux to rename files or directories with periods as backups but not to be processed, e.g. `file.conf.orig`, `file.rpmsave`, ... So we should also ignore apps with directory names that contain periods in them.

This would be beneficial if we went the RPM route and used symbolic links for the bundled OOD apps. Also this has affected Susan at PSC when she renamed an app to `jupyter_app.orig` as she was deploying an update and wanted to keep a backup of the original.

nickjer:
Fixes #313

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- Ignore apps if they have a period in directory name.
+  [#313](https://github.com/OSC/ood-dashboard/issues/313)
+
 ### Fixed
 - Fixed tests with class name collision.
 - Fixed tests reading global configuration when they shouldn't.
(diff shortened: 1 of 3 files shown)


ericfranz:
Whats the behavior for dev and shared apps?

nickjer:
It treats them as it does with hidden directories, and hides them from all the lists because you check if the apps have `#valid_dir?`:

- https://github.com/OSC/ood-dashboard/blob/343a9248ec2c6008eed4d22a3830c02116b15751/app/apps/sys_router.rb#L20
- https://github.com/OSC/ood-dashboard/blob/343a9248ec2c6008eed4d22a3830c02116b15751/app/apps/usr_router.rb#L45
- https://github.com/OSC/ood-dashboard/blob/343a9248ec2c6008eed4d22a3830c02116b15751/app/apps/dev_router.rb#L17

ericfranz:
If a manifest is missing, in production we hide the apps, and in development, we show the apps but have a nice warning stating that the app is missing a manifest. We should do the same here. If the app directory name has a . in it, show it but have the warning.

nickjer:
I now treat apps with periods the same as apps with invalid manifests. They don't show up in production, but they do show up in the Developer views with lint errors.

image

But the Dashboard has routing issues when dealing with these apps. Not sure if we should take the time to fix all these routing issues (e.g., can't edit the app from the Dashboard developer view or launch the app from the Dashboard).

PR #330, opened 2018-01-31
Update panel when time changes

### LINKED ISSUE #319: Update Interactive App panel if `time_used` changes
The panel doesn't update if the remaining time changes. A user will need to refresh the page to see how much time is left.

This can be easily updated by adding `info.wallclock_time` to the method `Session#to_hash` here:

https://github.com/OSC/ood-dashboard/blob/adec2529f8d4580487524ae65fe538dcc31ff0c4/app/models/batch_connect/session.rb#L401-L410

nickjer:
Fixes #319


ericfranz:
I do not understand this change. Why does this have any affect on panel updates:

```
time: info.wallclock_time.to_i / 60     # only update every minute
```

PR #332, opened 2018-02-02
Fix batch connect navigation

### LINKED ISSUE #173: Weird view if no system batch connect apps
Without system batch connect apps the views look odd:

screenshot from 2017-06-20 13-15-37

### LINKED ISSUE #259: If there are no valid apps, sub categories still show up
Install a generic batch connect app that specifies no cluster, or a cluster that does not exist (or any other reason that would make the app "invalid".

The app does not appear in the menu, but the category does:

screen 2017-10-09 at 1 30 56 pm

### LINKED ISSUE #261: Developer views should show batch connect "subapps" (if available)
clone bc_desktop
add a valid subapp
you cannot launch this from the developer views

### LINKED ISSUE #317: Add menu for Shared/Sandbox Interactive Apps
Currently the Interactive Sessions view only shows a menu of System Interactive Apps to the left. This makes navigating to a Shared or Sandbox Interactive App tedious.

A mock screenshot of what it could look like:

image

Although we may want this collapsible as some users have a lot of Interactive Apps in their Sandbox.

### LINKED ISSUE #318: Opening an Interactive App in Dashboard opens new tab
If you click and Interactive App from the Dashboard navbar it will open a new tab. This is not necessary as Interactive Apps are run within the Dashboard.

Although opening the form in a new tab may be useful for the Developer View's details page.

nickjer:
Fixes #173
Fixes #259
Fixes #317 
Fixes #318 
Fixes #261


app/views/apps/_app.html.erb
```diff
@@ -1,16 +1,43 @@
...
-    <div class="center-block" style="text-align: center">
-      <%= app_icon_tag(app) %>
-      <%= content_tag(:p, app.title) %>
-      <%= content_tag(:p, content_tag(:span, app.caption), class: "text-muted", style: "margin-top: 20px") if app.caption %>
+<%- if app.batch_connect_app? -%>
+  <%- app.batch_connect.sub_app_list.select(&:valid?).each do |sub_app| -%>
+    <div class="col-sm-3 col-md-3">
+      <%=
+        link_to(
+          new_batch_connect_session_context_path(token: sub_app.token),
```
ericfranz: So difference between these two is this in psuedocode:

```ruby
if app.batch_connect_app?
  app.batch_connect.sub_app_list.select(&:valid?).each do |sub_app|
    render_icon_view_for sub_app with sub_app.url
  end
else
  render_icon_view_for app with app.url
end
```

Two options to simplify:

1. create a partial for the actual icon and render that, retaining the logic here
2. add a new method `OodApp#valid_sub_apps` and for default returns `[self]` then you have one loop, using `#url` for the url

Since we have separate OodApp and BatchConnect::App objects methods might need to exist on both

nickjer: Doing a quick review of the methods called in those blocks, I would need:

- `BatchConnect::App#batch_connect_app?` (to detect whether to open in new tab)
- `BatchConnect::App#caption`
- `BatchConnect::App#icon_path`
- `BatchConnect::App#name`
- `BatchConnect::App#type`
- `BatchConnect::App#owner`
- `BatchConnect::App#manifest`
- `OodApp#description`

For each of the `BatchConnect::App` methods I would need to treat appropriately for sub-apps as well. So exposing `BatchConnect::App#manifest` may be ill-advised since the following may occur:

```ruby
my_sub_app.description == my_sub_app.manifest.description
# => false
```

So if I don't implement the method `BatchConnect::App#manifest` for safety reasons, then we need to add:

- `OodApp#description`
- `OodApp#icon`
- `BatchConnect::App#icon`

nickjer: I think another option is to use the `OodApp#links` method and build up the `OodApp::Link` object with more properties.

ericfranz:
I have terrible news... a design just occurred to me that may be sufficient to get rid of the interactive apps index page. Will share Monday. This exercise did result in the conclusion of needing a better data structure for apps though when dealing with a tree so that's good

app/views/apps/_app.html.erb
ericfranz: Reviewed the code changes and they look great

ericfranz:
Reviewed recent refactoring and it looks great

nickjer:
I missed your comment earlier when I was working on the code and pushed up the last commit. Just now saw it.

We can definitely meet on Monday to discuss your idea. Also we (mostly you, since I have no idea) need to think about what to do if there are no system interactive apps and the developer needs to access the Interactive Sessions view.

PR #340, opened 2018-02-09
Add interactive sessions option with responsive text

brianmcmichael:
This has some issues with the location of the collapse breakpoints.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -27,20 +27,8 @@
       <div class="collapse navbar-collapse" id="bs-example-navbar-collapse-9">
         <ul class="nav navbar-nav">
           <%= render partial: 'layouts/nav/group', collection: @nav_groups %>
-          <% if Configuration.app_development_enabled? || @nav_groups.has_batch_connect_apps? %>
-            <li title="My Interactive Sessions">
-              <a target="_self" href="<%= batch_connect_sessions_path %>">
-                <i class="fa fa-list"></i><span class="hidden-sm hidden-md hidden-md-nav"> My Interactive Sessions</span>
-              </a>
-            </li>
-          <% end %>
-          <% if Configuration.show_all_apps_link? %>
-            <li title="All Apps">
-              <a target="_self" href="<%= apps_index_path %>">
-                <i class="fa fa-th"></i><span class="hidden-sm hidden-md hidden-md-nav"> All Apps</span>
-              </a>
-            </li>
-          <% end %>
+          <%= render partial: 'layouts/nav/sessions', nav_groups: @nav_groups if Configuration.app_development_enabled? || nav_groups.has_batch_connect_apps? %>
+          <%= render partial: 'layouts/nav/all_apps' if Configuration.show_all_apps_link? %>
         </ul>
         <div class="navbar-right">
           <ul class="nav navbar-nav">
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -28,6 +28,13 @@
         <ul class="nav navbar-nav">
           <%= render partial: 'layouts/nav/group', collection: @nav_groups %>
           <%= nav_link("All Apps", "th", apps_index_path) if Configuration.show_all_apps_link? %>
+          <% if Configuration.app_development_enabled? || @nav_groups.has_batch_connect_apps? %>
+          <li>
+            <a target="_self" href="<%= batch_connect_sessions_path %>">
+              <i class="fa fa-th"></i><span class="hidden-sm hidden-md"> Interactive Sessions</span>
+            </a>
+          </li>
+          <% end %>
         </ul>
         <div class="navbar-right">
           <ul class="nav navbar-nav">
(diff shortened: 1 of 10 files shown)


brianmcmichael:
This adds the link to interactive sessions when 

`if(developer mode enabled OR a batch connect app exists in sys apps list){ show link }`

It also adds a couple of custom media queries to prevent navbar overflow.

app/views/layouts/nav/_develop_dropdown.html.erb
```diff
@@ -1,6 +1,6 @@
 <li class="dropdown" title="Develop">
   <a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button" aria-haspopup="true" aria-expanded="false">
-    <i class="fa fa-code"></i><span class="hidden-sm"> Develop</span><span class="caret"></span>
+    <i class="fa fa-code"></i><span id="develop-nav-text" class="hidden-sm"> Develop</span><span class="caret"></span>
```
ericfranz: If you want to do this, wouldn't `class="visible-lg-*"` be basically the same thing?

app/views/layouts/application.html.erb
```diff
@@ -28,6 +28,13 @@
         <ul class="nav navbar-nav">
           <%= render partial: 'layouts/nav/group', collection: @nav_groups %>
           <%= nav_link("All Apps", "th", apps_index_path) if Configuration.show_all_apps_link? %>
+          <% if Configuration.app_development_enabled? || @nav_groups.has_batch_connect_apps? %>
+          <li>
+            <a target="_self" href="<%= batch_connect_sessions_path %>">
+              <i class="fa fa-th"></i><span id="interactive-sessions-nav-text" class="hidden-sm hidden-md"> Interactive Sessions</span>
```
ericfranz: Instead of `interactive-sessions-nav-text` id, can't you achieve the same thing by just using class="visible-lg-*" or class="hidden-sm hidden-md hidden-lg".

app/views/layouts/nav/_develop_dropdown.html.erb
ericfranz: Misread, you were hiding this at 997px which is very close to 992px.

class ="visible-md-* visible-lg-*"

ericfranz: But actually instead of visible the hidden classes are better so you can keep the text visible at the smallest width.

app/views/layouts/application.html.erb
ericfranz: But actually instead of visible the hidden classes are better so you can keep the text visible at the smallest width.

ericfranz:
Also, this adds the link to the main nav bar. We should also remove the original link being added to the Interactive Apps dropdown (and all dropdowns that have a batch connect app).

brianmcmichael:
The default responsive utilities alone don't work. I can remove the custom queries but there will be a navbar overlap between 1200-1210px and 992-997px on my configuration.

app/views/layouts/nav/_develop_dropdown.html.erb
brianmcmichael: At the smallest width it collapses to the hamburger.

brianmcmichael: bootstrap "md" cuts off at 992, there is overlap between 992 and 997. 
bootstrap "lg" cuts off at 1200, there is overlap between 1200 and 1210.

ericfranz:
I was peeking at the responsive-utilities.scss file in bootstrap 3.

```scss
@media (max-width: $screen-xs-max) {
  @include responsive-invisibility('.hidden-xs');
}

@media (min-width: $screen-sm-min) and (max-width: $screen-sm-max) {
  @include responsive-invisibility('.hidden-sm');
}

@media (min-width: $screen-md-min) and (max-width: $screen-md-max) {
  @include responsive-invisibility('.hidden-md');
}

@media (min-width: $screen-lg-min) {
  @include responsive-invisibility('.hidden-lg');
}
```

You could add `hidden-md` to the "Develop". That will address one of the two problems. For the second

You could look at just adjusting the `$screen-lg-min` and corresponding `$screen-md-max` by 10 pixels.

Or 

Or add another mediaquery that adds a class to hide <=1300 pixels and show >=1300 pixels and use that class on the interactive apps and all apps.

I think these are better options than special css that targets id's of the elements directly. It isn't a requirement for the labels for "interactive apps" and "all apps" to ever actually appear, as there are recognizable icons that users can learn (and popovers could explain their purpose.

Finally, All Apps should after Interactive Apps, not before.

ericfranz:
One final change. From Alan's suggestion, rename "Interactive Sessions" to "My Interactive Sessions". That means changing the link and changing the breadcrumbs in the batch connect views.

```
NC166808% grep "Interactive Sessions" app -R
app/views/batch_connect/sessions/index.html.erb:<% content_for :title, "Interactive Sessions" %>
app/views/batch_connect/sessions/index.html.erb:    Interactive Sessions
app/views/batch_connect/session_contexts/new.html.erb:    <%= link_to "Interactive Sessions", batch_connect_sessions_path %>
app/views/layouts/application.html.erb:                <i class="fa fa-list"></i><span class="hidden-sm hidden-md hidden-md-nav"> Interactive Sessions</span>
NC166808% grep "Interactive Sessions" test -R
test/controllers/dashboard_controller_test.rb:      "Interactive Sessions",
```

PR #343, opened 2018-02-21
fix crash due to wrong nav_groups reference

### LINKED ISSUE #342: Dashboard crashes when development mode disabled
```
ActionView::Template::Error: undefined local variable or method `nav_groups'
```

ericfranz:
Fixes #342 

- the code was using `nav_groups` instead of `@nav_groups`
- also updated the tests to reflect the new "My Interactive Sessions" link


ericfranz:
@brianmcmichael please review

PR #346, opened 2018-02-26
Add Travis CI support

### LINKED ISSUE #345: Add Travis CI testing
Need to run tests in Travis CI.

nickjer:
Fixes #345


nickjer:
You might want to Squash & Merge this as I was experimenting with Travis CI.

test/controllers/dashboard_controller_test.rb
```diff
@@ -91,28 +90,34 @@ def dropdown_list_items(list)
...
 
-  test "should not create My Interactive Apps link" do
+  test "should not create My Interactive Apps link if no Interactive Apps and not developer" do
+    Configuration.stubs(:app_development_enabled?).returns(false)
     SysRouter.stubs(:base_path).returns(Rails.root.join("test/fixtures/sys"))
 
     get :index
-    assert css_select("li[title='My Interactive Sessions']").empty?
+    assert_response :success
+    assert_select "li[title='My Interactive Sessions']", count: 0
```
ericfranz: I was hoping that you could have done something like this instead:

```
assert_select "li[title='My Interactive Sessions'] a", "My Interactive Sessions", count: 0
```

Otherwise if the link's text remains i.e. "My Interactive Sessions" but we start omitting the "title" attribute, this test would never fail.

test/system/configuration_singleton_test.rb
```diff
@@ -41,10 +41,6 @@ def config_via_runner(env: 'development', envvars: '')
     Marshal.load(runner(code, env: env, envvars: envvars))
   end
 
-  test "tests should not be run with .env.local* files in directory" do
-    assert Dir.glob(".env.local{.development,.production,}").none?, "these tests should not be run with a .env.local or .env.local.development or .env.local.production"
-  end
-
```
ericfranz: Why remove this? This is valuable because sometimes when I run tests I don't realize I have .env.local files modifying the environment that the tests run in.

test/controllers/dashboard_controller_test.rb
nickjer: Maybe, but then should I remove the `[title='...']` then?

test/system/configuration_singleton_test.rb
nickjer: Because of...

https://github.com/OSC/ood-dashboard/blob/bb1f18d42d64b0d587deaa068ec0f06baba0aee7/config/configuration_singleton.rb#L143-L148

it will never run `.env.local`, `.env.development.local`, or `.env.production.local` in the test environment.

I also always have `.env.local` in my sandbox, and it is getting annoying to back it up, run tests, then move back.

ericfranz: Great!

PR #355, opened 2018-03-30
Various patches for announcements

### LINKED ISSUE #352: Cannot include HTML in announcment message
It seems I am using the Markdown renderer that is used for rendering the manifest of user apps. This is a more restrictive version meant to block HTML syntax for security purposes.

As the sys admin is the only one capable of generating an announcement read by the Dashboard (similar to the MOTD), it should use the more general Markdown renderer included in the AppKit. This will grant the sys admin more freedom when constructing the announcements to also include HTML.

### LINKED ISSUE #353: Don't render announcement message unless needed
Currently the announcement message is rendered whenever the `Announcement` object is initialized:

https://github.com/OSC/ood-dashboard/blob/c92b74ef16b849d795a0024cbc79fddc7ebcb461/app/models/announcement.rb#L25

As the rendering process *may be* expensive (e.g., calling the `quota` command can sometimes take >100 milliseconds) we should only render it when `Announcement#msg` is called instead (as before it should also be cached after rendering).

### LINKED ISSUE #354: No announcements display if one raises error
If parsing one of the announcement files raises an error (e.g., error in ERB code) then it fails to show any announcements. So valid announcements won't show up.

nickjer:
Fixes #352 
Fixes #353 
Fixes #354

### DIFF
--- a/app/models/announcement.rb
+++ b/app/models/announcement.rb
@@ -2,48 +2,32 @@ class Announcement
   # List of valid announcement types
   TYPES = %i(warning info success danger)
 
-  # The announcement's message
-  # @return [String, nil] the announcement's message if it exists
-  attr_reader :msg
-
-  # The type of announcement
-  # @return [Symbol] the type of announcement
-  attr_reader :type
-
-  class << self
-    # Parse a file
-    # @param path [#to_s] path to file
-    # @return [Announcement] announcement object
-    def parse(path)
-      path = Pathname.new(path.to_s).expand_path
-
-      if path.file? && path.readable?
-        case path.extname
-        when ".md"
-          Announcement.new(msg: path.read)
-        when ".yml"
-          Announcement.new(YAML.safe_load(ERB.new(path.read, nil, "-").result))
-        else
-          Announcement.new
-        end
-      else
-        Announcement.new
-      end
-    end
-  end
-
-  # @param opts [#to_h] the announcement object
+  # @param opts [#to_h, #to_s] the announcement object or path
   # @option opts [#to_sym] :type (:warning) Type of announcement (:info,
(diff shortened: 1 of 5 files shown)


app/models/announcement.rb
```diff
@@ -2,53 +2,53 @@ class Announcement
...
-  # @param opts [#to_h] the announcement object
+  # @param opts [#to_h, #to_s] the announcement object or path
   # @option opts [#to_sym] :type (:warning) Type of announcement (:info,
   #   :success, :warning, :danger)
   # @option opts [#to_s] :msg (nil) The announcement message
   def initialize(opts = {})
-    opts = opts.to_h.symbolize_keys.compact
+    if opts.respond_to? :to_h
+      @opts = opts.to_h
+    elsif opts.respond_to? :to_s
```
ericfranz: Every Ruby object responds to `to_s`. I think what you are looking for is `opts.respond_to? :to_str`.

nickjer: Unfortunately it needs to be `#to_s` because it could be initialized with a `Pathname` object. But as you point out all objects respond to `#to_s` in Ruby, so the final `else` statement is probably useless.

nickjer: An alternative is...

```ruby
if opts.respond_to? :to_h
  @opts = opts.to_h
elsif !opts.nil?
  @opts = Pathname.new(opts.to_s).expand_path
else
  @opts = {}
end
```

in the off chance someone supplies `nil` for the initialization parameter instead of leaving it blank.

nickjer: I also need to move the `#expand_path` down to the private `#opts` method in case that raises an error.

PR #359, opened 2018-04-19
Fix crash for invalid session data

### LINKED ISSUE #357: Better handling for batch_connect db errors
If an exception is raised when attempting to read from the database, for example parsing a file that has an invalid JSON format, it should handle this more gracefully than displaying a stack trace to the user.

image

This is related to S/N ticket: [INC0327952](https://oarnet.service-now.com/nav_to.do?uri=incident.do?sys_id=a9f5e0fadb55d3c0321f1aac0b96194d)

nickjer:
Fixes #357


app/models/batch_connect/session.rb
```diff
@@ -77,8 +77,14 @@ def db_root
       # @return [Array<Session>] list of sessions
       def all
         db_root.children.select(&:file?).map do |f|
-          new.from_json(f.read)
-        end.map do |s|
+          begin
+            new.from_json(f.read)
+          rescue => e
+            Rails.logger.error("ERROR: Error parsing file '#{f}' --- #{e.class} - #{e.message}")
+            f.delete unless Rails.env.development?
```
ericfranz: Auto clean up may be nice but could be annoying in production for sites debugging problems as a corrupted session file would disappear. When can a session file become corrupted to produce this error?

nickjer: If it isn't cleaned up then logs will become full of error messages for every corrupt file.

> When can a session file become corrupted to produce this error?

Right now only if the Linux kernel is interrupted during the write process (e.g., due to quota limit being reached or power outage). Or if the user goes in there and edits the files by hand.

app/models/batch_connect/session.rb
```diff
@@ -76,9 +76,15 @@ def db_root
       # Find all active session jobs
       # @return [Array<Session>] list of sessions
       def all
-        db_root.children.select(&:file?).map do |f|
-          new.from_json(f.read)
-        end.map do |s|
+        db_root.children.select(&:file?).reject {|p| p.extname == ".bak"}.map do |f|
```
ericfranz: I believe that `db_root.children` crashes if `db_root` doesn't exist or isn't a directory or is a directory but isn't executable. I ran into a similar problem with the routers:

https://github.com/OSC/ood-dashboard/commit/8c92bff63d21a15d549566f9436f02513fa76ebc

A guard clause at the top could be as simple as:

```ruby
return [] unless db_root.directory? && db_root.executable? && db_root.readable?
```

ericfranz: After protecting against this crash, everything else looks good, assuming it works :-).  No need for me to review again.

nickjer: If it doesn't exist it auto-creates it. Although I do not check if it is a directory that a user has read, write, and execute access to. That can get a little messy to handle all of those situations as `db_root` is called in multiple places and I would need to propagate the errors all the way up to the user in the views.

nickjer: I opened up the issue https://github.com/OSC/ood-dashboard/issues/361 to address the concerns brought up here as it would require its own requirements gathering, design, and PR.

PR #367, opened 2018-05-24
Support disk quotas and add warnings

### LINKED ISSUE #366: Add quota limit warnings
Hitting the quota limit has affected multiple users (commercial and academic) and has led to web apps breaking. In particular, for the case of one of our commercial clients it caused his database to become corrupt. It usually results in tickets or people stopping by our cubicle asking why an app like Job Composer doesn't work and about half an hour of our time to figure out that it is due to the user's quota.

We would like to be able to warn a user if they are nearing their quota limit (disk space and number of files) for all volumes the user has quota restrictions on and is utilizing disk resources under.

nickjer:
Fixes #366 

See the `README.md` for configuration options.

The below figure was generated with a `.env.local` set as:

```shell
OOD_QUOTA_PATH="/users/reporting/storage/quota/netapp.netapp-home.ten.osc.edu-users_quota.json:/users/reporting/storage/quota/gpfs.project_quota.json"
OOD_QUOTA_THRESHOLD="0.39"
```

image


app/presenters/quota_presenter.rb
```diff
@@ -0,0 +1,92 @@
+# Decorator for {Quota} object making it more presentable for the view
+class QuotaPresenter < SimpleDelegator
+  # Size of block in bytes
```
ericfranz: I think that many of these methods belong on Quota instead of the presenter.

For example:

```
block_usage_in_bytes
total_block_usage_in_bytes
block_limit_in_bytes
total_block_usage_percent
total_file_usage_percent
```

Those legitimately could be on the Quota object itself. If those are moved to Quota, including BLOCK_SIZE then I question the need for this class.

For example in erb:

```ruby
quota.human_block_usage
```

is not a huge difference from 

```ruby
number_to_human_size(quota.block_usage_in_bytes)
```

though it does look like some of the methods below (human_resource_limit and total_resource_usage_percent) provide a helpful abstraction to keep the view really simple.

app/decorators/sufficient_quota_decorator.rb
```diff
@@ -0,0 +1,43 @@
...
+  attr_reader :threshold
+
+  # @param quota [Quota] the quota object we want to decorate
+  # @param opts [#to_h] initialization options
+  # @option opts [#to_f] :threshold threshold percentage
+  def initialize(quota, opts = {})
+    super(quota)
+
+    opts = opts.to_h.compact.symbolize_keys
+    @threshold = opts.fetch(:threshold, Configuration.quota_threshold).to_f
```
ericfranz: Configuration.quota_threshold should not be in this class. We only instantiate SufficientQuotaDecorator in one place, so I would do this:

```
SufficientQuotaDecorator.new(quota, threshold: Configuration.quota_threshold)
```

I'm certain I have done this same thing in other areas of the code and I should fix those.

app/models/quota.rb
```diff
@@ -0,0 +1,112 @@
...
+      paths = Array.wrap(paths).map { |p| Pathname.new p }
+
+      quotas = []
+      paths.each do |path|
+        json = Oj.strict_load(path.read)
+        case json["version"].to_i
+        when 0..1
+          quotas += find_v1(user, json)
+        else
+          Rails.logger.error("Invalid version specified for quota file: '#{path}'")
```
ericfranz: Maybe the logger should be passed in as an argument as well: `opts.fetch(:logger)`

app/models/quota.rb
```diff
@@ -0,0 +1,112 @@
...
+    # Get quota objects only for requested user in JSON file(s)
+    def find(user, opts = {})
+      user  = user && user.to_s
+      opts  = opts.to_h.compact.symbolize_keys
+      paths = opts.fetch(:paths, Configuration.quota_paths)
+      paths = Array.wrap(paths).map { |p| Pathname.new p }
+
+      quotas = []
+      paths.each do |path|
+        json = Oj.strict_load(path.read)
```
ericfranz: In your tests, `Oj` was on avg 29ms vs regular `JSON` which was avg 47ms. 19ms 
difference. Not sure that this is worth adding another third party dependency.

Though for non-core OOD apps, both this dependency and https://github.com/ohler55/ox look interesting.

ericfranz:
For completeness and compatibility with other centers, we should support `Quota#file_grace` and `Quota#disk_grace`. We currently don't use that feature right now. Might open an issue after pulling this in.

ericfranz:
Its worth consideration: there is another way to model this object besides the existing Quota object which corresponds to an array element in the json file.

A more targeted Quota object would deal with a single file or block quota:

```ruby
class Quota
  attr_reader :path, :user, :type, :limit, :user_usage, :total_usage, :unit

  def total_usage_percentage
  end

  def sufficient?(threshold: 0.85)
  end

  def insufficient?
  end
end
```

1. Since we only report in bytes, we can take a Quota instance in blocks and actually create a new Quota instance in bytes
2. `sufficient?` can just be a method on the object that accepts a %

The only problem with this approach is that it appears we would need to implement the humanize methods on this class (or relevant subclasses) unless we were going to add some ugly "if unit == X call number_to_human_size else call number_to_human" in the view (or a helper.

This removes all the duplicate methods for "files" and "blocks" and would simplify the views as well. Not sure if there is enough time to experiment with this approach to actually see if it is worth the tradeoffs.

laser505t:
How universal are these JSON files at other HPC centers? What might there need be done to the Quota class to increase compatibility? Is 1 KB always the block size in a given Linux HPC system? How would the grace files and blocks show up in the JSON file?
Should I go ahead and change Oj.strict_load() to JSON.parse() and remove that gem?
Should we put this warning in the sandbox apps listing too? Might make sense so a user would be inclined to delete some of those that haven't been used in a while.
I guess how we model the Quota class depends on whether we want it to be based on one path, or more specific with the type of quota resource. Kinda goes back to what quotas are like at other centers.

PR #368, opened 2018-06-01
Change clone app form to indicate https cloning

### LINKED ISSUE #256: Inform user that git http urls also work when cloning apps
New Product form suggests only git ssh urls work in the web form. The Git remote example shows an ssh url, claiming it requires SSH key installed, but does not provide an HTTPS example, which does not require SSH key.

A big alert appears with the "Generate SSH Key", without indicating what commands will be run as the user when clicking "Generate SSH Key". This is not helpful if I'm using https urls.

- [ ] 1. MVP: Provide https example in help, and indicate no SSH key is required for this
- [ ] 2. Optionally do not display "No SSH key found..." alert unless the URL inserted is an SSH URL (maybe vue.js could help with this?)

laser505t:
Fixes #256

### DIFF
--- a/app/views/products/_form_git.html.erb
+++ b/app/views/products/_form_git.html.erb
@@ -1,5 +1,5 @@
 <div class="field">
-  <%= f.text_field :git_remote, help: "Example: <strong>git@github.com:Example/example.git</strong> (<a href='https://help.github.com/articles/generating-an-ssh-key/' target='_blank'>requires SSH key installed</a>)".html_safe %>
+  <%= f.text_field :git_remote, help: "Example: <strong>git@github.com:Example/example.git</strong> (<a href='https://help.github.com/articles/generating-an-ssh-key/' target='_blank'>requires SSH key installed</a>) or <strong>https://github.com/Example/</strong>".html_safe %>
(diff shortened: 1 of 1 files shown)


app/views/products/_form_git.html.erb
```diff
@@ -1,5 +1,5 @@
 <div class="field">
-  <%= f.text_field :git_remote, help: "Example: <strong>git@github.com:Example/example.git</strong> (<a href='https://help.github.com/articles/generating-an-ssh-key/' target='_blank'>requires SSH key installed</a>)".html_safe %>
+  <%= f.text_field :git_remote, help: "Example: <strong>git@github.com:Example/example.git</strong> (<a href='https://help.github.com/articles/generating-an-ssh-key/' target='_blank'>requires SSH key installed</a>) or <strong>https://github.com/Example/</strong>".html_safe %>
```
ericfranz: If the example for ssh is `git@github.com:Example/example.git` then the example for https would likely be `https://github.com/Example/example.git` rather than `https://github.com/Example/`.

i.e.  git@github.com:OSC/ood-dashboard.git and https://github.com/OSC/ood-dashboard.git

PR #370, opened 2018-07-24
Upgrade to FontAwesome 5

### LINKED ISSUE #341: Upgrade Font Awesome to 5
Font Awesome is now up to version 5, and the SASS gem has recently been updated as well:

https://github.com/FortAwesome/font-awesome-sass

This does introduce some breaking changes as some of the icon names have changed and there is a new icon style that needs to be specified *solid*, *regular*, and *brands*.

laser505t:
Fixes #341 

Apps now have the ability to specify a FontAwesome style in the icon URI in their manifest.yml. For apps that still use 'fa://', the shims stylesheet will take care of it. However, the style MUST be specified in the URI if using the newer names. For example, if a user wants a gear icon and wants to use the FontAwesome 5 name for it, which used to be 'gear' but is now 'cog', 'fas://cog' must be used as opposed to just 'fa://cog'. However, 'fa://gear' will still work. In other words, 'fa://' is only for backward compatibility and should not be used on newly-created apps. We should update the manifests of the core apps and example apps to reflect this.


app/helpers/application_helper.rb
```diff
@@ -39,25 +39,40 @@ def passwd_url
     ENV['OOD_DASHBOARD_PASSWD_URL']
   end
 
-  def fa_icon(icon)
-    content_tag(:i, "", class: ["fa", "fa-#{icon}", "fa-fw", "app-icon"] , title: "FontAwesome icon specified: #{icon}")
+  def fa_icon(icon, fa_style = "fas")
```
ericfranz: better to use keyword args here:

```ruby
def fa_icon(icon, fa_style: "fas")
...
```

app/helpers/application_helper.rb
```diff
@@ -39,25 +39,40 @@ def passwd_url
...
+    content_tag(:i, "", class: [fa_style, "fa-#{icon}", "fa-fw", "app-icon"] , title: "FontAwesome icon specified: #{icon}")
   end
 
   def app_icon_tag(app)
     if app.icon_path.file?
       image_tag app_icon_path(app.name, app.type, app.owner), class: 'app-icon', title: app.icon_path
     else # default to font awesome icon
-      icon = (app.manifest.icon =~ /fa:\/\/(.*)/) ? $1 : "gear"
-      fa_icon(icon)
+      if app.manifest.icon =~ /^fa[bsrl]?:\/\/(.*)/
```
ericfranz: you can wrap both matches in one regex:

```ruby
/^(fa[bsrl]?):\/\/(.*)/

$1 # == "fa"
$2 # == "icon"
```

Also this is interesting. I'd prefer to make the default case easiest. For example, in the future, if I did `fa://cog` I would like to see that translate to the class `fas fa-cog`. Because in the default case I just want to choose an icon, and not think about the icon style. In fact, we could stay we do not support different icon styles, only the "solid style".

But I gather that would actually be difficult because then you would need some logic to determine which icons are safe to use the old style? Does the shim make every single icon work with the class format "fas fa-ICON" such as "fas fa-gear"?

ericfranz: Or, if `fa://` is used then we default to `fas` for the style... is that possible? I guess it depends on how the shims work.

app/helpers/application_helper.rb
```diff
@@ -39,25 +39,40 @@ def passwd_url
...
-      fa_icon(icon_uri.host)
+      fa_icon(icon_uri.host, "fa")
+    when "fas"
+      fa_icon(icon_uri.host, "fas")
+    when "far"
+      fa_icon(icon_uri.host, "far")
+    when "fab"
+      fa_icon(icon_uri.host, "fab")
+    when "fal"
+      fa_icon(icon_uri.host, "fal")
```
ericfranz:
```ruby
if %w(fa fas far fab fal).include?(icon_uri.scheme)
  fa_icon(icon_uri.host, icon_uri.scheme)
else
  image_tag icon_uri.to_s, class: "app-icon", title: icon_uri.to_s
end
```

app/helpers/application_helper.rb
laser505t: `fas fa-gear` does not work; it must be `fa fa-gear`. That is why I think we should move on to changing the way the URI works and leave `fa://` only for backward compatibility.

ericfranz: Oh but according to https://fontawesome.com/how-to-use/on-the-web/setup/upgrading-from-version-4 `fa fa-cog` should work i.e. `Font Awesome Solid == fas or fa`. Furthermore, `far` and `fal` will only work if we have the pro version, which we do not (cost $$).`fab` would work though.

Can you confirm that all the icons you changed to `fas` would work if just using `fa`?

laser505t: Actually, there are some free icons under `far`, but none for `fal`. But it does work if you use the new name for the icon and the old `fa` class. So default everything to just `fa`?

PR #371, opened 2018-07-24
Display a human-readable title for paths in Files menu

### LINKED ISSUE #156: Add optional title to "favorite paths" for the dashboard Files menu
For OSC it has worked out okay that the word "project" and "scratch" appear in the paths.

But at other sites this might not be the case. For example, at PSC my project space path is /pylon2/sy560jp/efranz and my scratch space path is /pylon5/sy560jp/efranz and would show up in the menu like:

screen shot 2017-05-23 at 3 59 40 pm

What would be preferable is if we could have human friendly titles, such as Project Space and Scratch Space:

screen shot 2017-05-23 at 4 01 55 pm

HTML looks like:

```html
<li>
  <a target="_blank" href="/pun/sys/files/fs/pylon2/sy560jp/efranz">
    <i class="fa fa-folder"></i> Project Space <br><small>/pylon2/sy560jp/efranz</small>
  </a>
</li>
```

laser505t:
Fixes #156 

I tried to incorporate it into the existing code in the most sensible way. I'm not sure how the <li class="divider"> element got in the menu in the screen clip in #156, but I deemed it too tricky to implement for this occasion. The reason I didn't make a new class like suggested is because this Files menu along with Jobs, Clusters, and Interactive Apps are all treated the same. They use the same Rails partial. They are for presenting the user with all the apps available in /var/www/ood/apps/sys/ in a nice way. The project and scratch spaces only show up in that menu because of a Rails initializer that isn't active when you do an install of the Dashboard straight from GitHub. Maybe we want to treat the Files menu separately?

### DIFF
--- a/app/apps/ood_app.rb
+++ b/app/apps/ood_app.rb
@@ -61,6 +61,7 @@ class OodApp
       file_links = [
         OodAppLink.new(
           title: "Home Directory",
+          subtitle: Dir.home.to_s,
           description: manifest.description,
           url: OodAppkit.files.url(path: Dir.home),
           icon_uri: "fa://home",
--- a/app/apps/ood_app.rb
+++ b/app/apps/ood_app.rb
@@ -58,27 +58,29 @@ class OodApp
   def links
     if role == "files"
       # assumes Home Directory is primary...
-      [
+      file_links = [
         OodAppLink.new(
           title: "Home Directory",
           description: manifest.description,
           url: OodAppkit.files.url(path: Dir.home),
-          icon_uri: icon_uri,
+          icon_uri: "fa://home",
           caption: caption,
           new_tab: true
         )
-      ].concat(
-        OodFilesApp.new.favorite_paths.map do |path|
+      ]
+      OodFilesApp.new.favorite_paths.each { |path, description|
+        file_links.concat([
           OodAppLink.new(
-            title: path.to_s,
+            title: description.to_s,
+            subtitle: path.to_s,
             description: manifest.description,
             url: OodAppkit.files.url(path: path),
             icon_uri: "fa://folder",
             caption: caption,
-            new_tab: true
-          )
-        end
-      )
+            new_tab: true)
+        ])
+      }
+      return file_links
     elsif role == "shell"
       login_clusters = OodCore::Clusters.new(
         OodAppkit.clusters
@@ -246,7 +248,7 @@ class OodApp
   # @return [Boolean] true if Gemfile.lock has specified gem name
   def has_gem?(gemname)
     # FIXME: we want to make this public, test it, and add functionality to make it
-    # work whether the app has a Gemfile.lock or just a Gemfile. 
+    # work whether the app has a Gemfile.lock or just a Gemfile.
     # see ood_app_test.rb
     has_gemfile? && bundler_helper.has_gem?(gemname)
   end
--- a/app/apps/ood_app_link.rb
+++ b/app/apps/ood_app_link.rb
@@ -1,5 +1,6 @@
 class OodAppLink
   attr_reader :title
+  attr_reader :subtitle
   attr_reader :description
   attr_reader :url
   attr_reader :icon_uri
@@ -9,6 +10,7 @@ class OodAppLink
     config = config.to_h.compact.symbolize_keys
 
     @title       = config.fetch(:title, "No title set").to_s
+    @subtitle    = config.fetch(:subtitle, "").to_s
     @description = config.fetch(:description, "").to_s
     @url         = config.fetch(:url, "").to_s
     @icon_uri    = URI(config.fetch(:icon_uri, "fa://gear").to_s)
--- a/app/apps/ood_files_app.rb
+++ b/app/apps/ood_files_app.rb
@@ -1,25 +1,26 @@
 # utility class for dealing with apps that play "files" role
 class OodFilesApp
   class << self
-    # an array of Pathname objects to check for the existence of and access to
-    # should always be []
-    # set in config/initializers/ood.rb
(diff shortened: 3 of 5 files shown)


ericfranz:
Sorry I didn't give good guidance on this issue. Part of the reason I suggested a class was because we could use ducktyping in the array. `config/examples/osc/initializers/ood.rb` needs to still work as it did before, without being changed. Sites external to OSC have already added to their initializers things like:

```ruby
OodFilesApp.candidate_favorite_paths.tap do |paths|
  # add project space directories
  projects = User.new.groups.map(&:name).grep(/^P./)
  paths.concat projects.map { |p| Pathname.new("/fs/project/#{p}")  }
 
  # add scratch space directories
  paths << Pathname.new("/fs/scratch/#{User.new.name}")
  paths.concat projects.map { |p| Pathname.new("/fs/scratch/#{p}")  }
end
```

So its expected at least that this array will contain Pathname objects. `candidate_favorite_paths` needs to stay an array and cannot be converted to a hash. But to support paths with "titles" we now need to add a new object that can contain both a path and a title. So the options for that are:

1. Create a well defined class that contains both a Pathname and a Title as attributes, and the OodFilesApp class that uses this list needs to coerce the resulting objects into the same object type before having it be processed by the view
2. Use a decorator class that adds title to the pathname, but delegates the rest of the method calls to the original Pathname object. All the code assumes it is just a Pathname object except the view where it can display the optional title if the object `respond_to? :title`
3. Something perhaps more complex? such as a separate hash that contains path => title lookup. It could be stored as an instance of the `OodFilesApp` and we could add a helper OodFilesApp.add_candidate_favorite_path(pathname, title=nil) then OodFilesApp.titles returns that hash i.e.

```ruby
def self.titles
  @titles ||= {}
end
```

And usage: `OodFilesApp.titles[some_pathname_obj] = "Scratch Space"` and `OodFilesApp.titles.fetch(some_pathname_obj, "")`.

Though that is not that object oriented and moves logic into the views.

ericfranz:
1. OodFilesApp.candidate_favorite_paths should be an array that can have in it strings, pathname objects, or FavoritePath class... or maybe even a hash with {path:, title:} (is that another suggestion I said?)
2. This method should be responsible for coercing the various objects from the class attribute self.class.candidate_favorite_paths into an array of common objects. https://github.com/OSC/ood-dashboard/blob/b2ecdd5dffdb8334a91839aa3216be1b91dda5f5/app/apps/ood_files_app.rb#L13-L15. It could use a private method to handle this. Instead of being pathnames, it probably should be a class that contains a pathname.
3. That class FavoritePath could have a method "accessible?" that checks it is accessible by the user so you could replace:

https://github.com/OSC/ood-dashboard/blob/b2ecdd5dffdb8334a91839aa3216be1b91dda5f5/app/apps/ood_files_app.rb#L20-L24

with

```ruby
  def favorite_paths
    @favorite_paths ||= candidate_favorite_paths.select(:accessible?)
  end
```

laser505t:
This allows the user configuring the favorite paths to use one of three things:
An array of Pathnames
An array of FavoritePaths
An array of hashes, where the hashes consist of {Pathname => <String> title} and can be more than one key-value pair

These three will be coerced into a simple array of FavoritePaths.
This has been tested with the following three initializers:
```
# array of hashes
OodFilesApp.candidate_favorite_paths.tap do |paths|
  # add project space directories (individual hashes)
  projects = User.new.groups.map(&:name).grep(/^P./)
  projects.each { |proj|
    paths << {Pathname.new("/fs/project/#{proj}") => "Project Space"}
  }

  # add scratch space directories (one big hash)
  tmp_scrt_hsh = {Pathname.new("/fs/scratch/#{User.new.name}") => "Scratch Space"}
  projects.each { |proj|
    tmp_scrt_hsh[Pathname.new("/fs/scratch/#{proj}")] = "Scratch Space"
  }
  paths << tmp_scrt_hsh
end
```

```
# FavoritePaths
OodFilesApp.candidate_favorite_paths.tap do |paths|
  # add project space directories
  projects = User.new.groups.map(&:name).grep(/^P./)
  paths.concat projects.map { |p| FavoritePath.new("/fs/project/#{p}", title: "Project Space")  }

  # add scratch space directories
  paths << Pathname.new("/fs/scratch/#{User.new.name}")
  paths.concat projects.map { |p| FavoritePath.new("/fs/scratch/#{p}", title: "Scratch Space")  }
end
```

```
# Pathnames
OodFilesApp.candidate_favorite_paths.tap do |paths|
  # add project space directories
  projects = User.new.groups.map(&:name).grep(/^P./)
  paths.concat projects.map { |p| Pathname.new("/fs/project/#{p}")  }

  # add scratch space directories
  paths << Pathname.new("/fs/scratch/#{User.new.name}")
  paths.concat projects.map { |p| Pathname.new("/fs/scratch/#{p}")  }
end
```

Maybe FavoritePath should be redesigned a little though, to instead have a path attribute instead of its path being accessed by to_s?
Is it good to show the path of the home directory? If so, I can change the test for Travis. If not, it's easy to revert.

laser505t:
The changes I made cause the description in the manifest.yml of ood-fileexplorer to be the title attribute of the nav link (what shows when you hover over it)
image
Before, the title attribute would be 'Home Directory', and '/fs/scratch/PZS0705' for the third link in the screen snip. I think that's a little redundant, and showing the manifest description matches other parts of the top menu (for example, Job Composer has 'Job Composer allows the creation and submission of batch jobs' for its title attribute).
This change can be reverted easily if necessary.

kcgthb:
Hi!
I was wondering if this feature was planned to be included in one of the future release?

ericfranz:
@kcgthb thank you for the ping, this dropped off my radar. I'll make sure it gets in 1.7.

PR #372, opened 2018-07-24
WIP: Show Readme in development app details page

### LINKED ISSUE #216: Display README in app details view
i.e. /show product

* display dependencies that the app depends on (rendered markdown or plain text - but render markdown only if extension is `md` or `markdown`) so that example apps copied state their dependencies clearly

laser505t:
Fixes #216 

It works, but the markdown renderer used makes the Readme look like crap. It even messes up on some. And it looks quite different from GitHub. I will look into whether there is another markdown renderer in the existing gems, otherwise look into what GitHub uses.


laser505t:
Here are some things that I found distasteful and/or problematic with how it is rendered:
	activejobs as example: numbering results in a bunch of 1s, but works when not tabbed over (bc_example_jupyter, very not good)
	no syntax highlighting (not important?)
	no <hr /> under titles (not important?)
	table of contents links don't work (ood-fileeditor)
	links go to same tab
	no images (ood-fileeditor)
	code blocks under subheading still take up full width (si-18-sinatra-base)
	Stuff that is tabbed over doesn't render that way (si-18-sinatra-base)
Which of these should be looked into to fix? Should it be changed in OodAppkit, or should I create a custom renderer just for the Readmes?

app/views/products/show.html.erb
```diff
@@ -82,7 +82,24 @@
 
 </div><!-- /.row -->
 
+<% if @product.readme? %>
+<hr />
+<div class="container row">
+  <button class="btn" data-toggle="collapse" data-target=".readme">Toggle Readme</button>
```
ericfranz: 1. Instead of a toggle, just show the contents of the README. Add a header above with the title the name of the readme, using HTML like: <div class="page-header"><h3>README.md</h3></div>
2. Follow the bootstrap guidelines: a div that is a row should always include a div that is a col. I think that "container" class is superfluous here.
3. That col should have the offset to the left so that the contents of the README line up with the rest of the attributes  (see bottom of comment)
4. Check and verify apps.awesim.org, especially on the "Shared apps" where the Permissions controls appear underneath the metadata. The permissions controls should appear above the README. In fact the permission controls should probably have the same offset as the readme and the metadata.


```html
<div class="row">
    <div class="col-md-offset-2 col-md-10 col-sm-10 col-xs-12"> <!-- col-md-6 or col-md-8 or col-md-10 -->
      <div class="page-header"><h3>README.md</h3></div>
```

screen 2018-07-31 at 3 33 55 pm
screen 2018-07-31 at 3 33 41 pm

laser505t:
The code blocks under ordered lists that make the list stay at 1 can be fixed by changing the tab spacing to 4 instead of 3 like we have in at least one Readme on GitHub.

laser505t:
Here are some but surely not all the repos that should have revised tab spacing:
ood-documentation
ood-myjobs
ood-dashboard
ood-activejobs
ood-example-batch

I still have yet to test this with app permissions showing.

app/views/products/show.html.erb
```diff
@@ -82,25 +82,10 @@
...
-
-<hr />
-
 <% if @product.permissions? %>
-
-<h3>Permissions</h3>
-
 <div class="row">
-  <div class="col-sm-6">
+  <div class="col-md-offset-2 col-md-10 col-sm-12 col-xs-12"><div class="page-header"><h3>Permissions</h3></div></div>
```
ericfranz: Format into multiple lines i.e.

```html
<div class="col-md-offset-2 col-md-10 col-sm-12 col-xs-12">
  <div class="page-header">
```

etc.

app/helpers/products_helper.rb
```diff
@@ -1,4 +1,73 @@
 module ProductsHelper
+
+  class RenderReadmeMarkdown < Redcarpet::Render::HTML
```
ericfranz: This should be moved to a separate class. Probably under another directory under app/. Not sure which yet. Until then could move to app/models/ and call the class ProductReadmeMarkdownRenderer. FWIW I don't know why the other renderer is called `RenderManifestMarkdown` which is a verb, not a noun.

app/models/product.rb
```diff
@@ -216,6 +216,21 @@ def uncommitted_files
...
+        @readme ||= path
+        found_readme = true
+      end
+    end
+    return found_readme
+  end
+
+  def readme
+    return readme? ? @readme : nil
+  end
```
ericfranz: This logic and the view logic for rendering the readme would be simpler if there were a separate `ProductReadme` class with methods `html`, `exist?`, `title`, `name`. You can also use the Rails view link and path/url helpers in the class by `include Rails.application.routes.url_helpers` so then you could add a `ProductReadme#edit_link`.

The method `ProductReadme#html` could return the rendered content, whether it is a plain text file `<pre>This README is not markdown formatted</pre>` or a markdown file `<h1>Markdown file</h1> <p>Yeah!</p>`.

For finding the readme you can use `Pathname.glob`:

```
2.2.0 :023 > app
 => #<Pathname:/Users/efranz/dev/ood-dashboard>
2.2.0 :024 > Pathname.glob app.join("readme*"), File::FNM_CASEFOLD
 => [#<Pathname:/Users/efranz/dev/ood-dashboard/README.md>]
```

Since glob always returns an array, you can do this:

```ruby
# returns
def readme_path
  Pathname.glob(app.join("readme*"), File::FNM_CASEFOLD).first
end
```

Of course it returns nil if there is no readme, which is not ideal but is okay in this case if the method is just used internally.

Also, of course, in the view you would still access the readme the same way: `@product.readme`

laser505t:
@ericfranz What was your idea for a 'name' method? I think the implementation I just did for 'edit_link' is good because it is generic and can be used elsewhere nicely, but maybe you wanted something that included all the heading formatting and whatnot.

laser505t:
Do we maybe want to allow configuration to turn this feature off with an environment variable?

ericfranz:
`ProductReadme#edit_link` currently is a misnomer - edit_url or edit_path would be more accurate (link in Rails usually means the entire HTML anchor `<a href=""...>` that the `link_to` helper produces)

The way you have it is good though: `ProductReadme#edit_url` as its own path is nice

I would however add a `ProductReadme#edit_link` maybe:

```ruby
#TODO: make sure to include Rails.application.routes.url_helpers at top
def edit_link(opts)
  link_to title, edit_url.to_s, title: "Edit #{title}", **opts 
end

# then in view
<h3>
  <%= product.readme.edit_link target: '_blank', style: "color: inherit; text-decoration: inherit;" %>
</h3>
```

I think thats what I was thinking. But If you rename edit_link to edit_url I would merge it as is as well.

PR #379, opened 2018-07-26
Refactor code and allow for more portability

ericfranz:
Incorporate SufficientQuotaDecorator into QuotaPresenter
Minimize dependency on Rails configuration
Rename Quota attributes for programmer clarity


app/models/quota.rb
```diff
@@ -40,18 +40,26 @@ class Quota
...
+
   class << self
     # Get quota objects for all users in JSON file(s)
     def all(opts = {})
       find(nil, opts)
     end
 
     # Get quota objects only for requested user in JSON file(s)
-    def find(user, opts = {})
+    def find(quota_paths, user, opts = {})
```
ericfranz: This method can be simplified further, by reducing its scope.

1. Instead of accepting multiple paths, just do 1 path
2. Instead of logging, raise an exception (custom named exception - here is an example of how to create custom exceptions https://github.com/OSC/ood_core/blob/396ec9e8720ac680fb11947e0e1e68b73c67ccd2/lib/ood_core/errors.rb)
3. No need for the opts hash, its not used anywhere so the method sig is now `find(quota_path, user)`

You need to change the one place this method is called, to call it on each path available and if an exception is raised, log it. And the result, two arrays, will need to be combined. But that doesn't take much code and can happen in one place outside of this class.

app/controllers/application_controller.rb
```diff
@@ -53,10 +53,8 @@ def set_announcements
   # an insufficient disk resource
   def set_my_quotas
     @my_quotas =
-      Quota.find(OodSupport::User.new).map do |quota|
-        SufficientQuotaDecorator.new(quota)
-      end.map do |quota|
-        QuotaPresenter.new(quota)
+      Quota.find(::Configuration.quota_paths, OodSupport::User.new).map do |quota|
+        QuotaPresenter.new(quota, threshold: ::Configuration.quota_threshold)
```
ericfranz: See this comment: https://github.com/OSC/ood-dashboard/pull/367/files#r193222458
One of the reasons this code is complex and has such a presenter is because the view wants to actually display several quota messages for displaying disk size and file size.

So we have this duplication all over the place. A final refactor stab could actually take a look at simplifying Quota class further, by instead of having both files and disk and percent total messages on files and disk, just having something like:

```ruby
class Quota
  attr_accessor :type, :resource, :user_usage, :total_usage, :units
end
```

More objects would be created (2 for each json hash instead of 1) but you could probably get by without a presenter then. If you do an exploration with this, I'd save it for last and do it on a separate (yet another) branch.

ericfranz:
I realized now that the presenter calls super so it wants to have the same initializer as the other class, thus this opts hash :-(.

app/controllers/application_controller.rb
ericfranz: Actually for this idea, it might also work to have `Quota`, `FileQuota < Quota`, `BlockQuota < Quota`.

app/views/shared/_insufficient_quota_resource.html.erb
```diff
@@ -1,31 +1,46 @@
+<% case quota.resource_type; when "file"; resource_title = "number of files"; when "block"; resource_title = "disk size"; end %>
```
ericfranz: Simplify by moving logic to Quota as discussed

i.e.


```ruby
def to_s
  if resource_type == "file"
    "Using #{number_to_human()} of quota #{number_to_human()} files."
  else
    "Using #{number_to_human_size()} of quota #{number_to_human_size()} files."
  end
end

# and
include ActionView::Helpers::NumberHelper
# for number_to_human_size & number_to_human
```

PR #381, opened 2018-08-02
Add icon picker to Edit Metadata

### LINKED ISSUE #380: Add font-awesome icon picker to app development edit metadata page
Can use this: https://farbelous.io/fontawesome-iconpicker but will have to add the old icons. It uses a metadata file for the new icons, so there may be a similar metadata file in a late version of Font AweSim 4 (to build the list of aliases).

It needs some design. Please do a mock design first and attach the image to this issue before proceeding!

1. If the icon.png file is present, it should show this icon as the current icon.png and say "if you want to use font awesome icon, first remove the icon.png from the root of the directory.
2. If no icon.png is present (that means font awesome icon is being used) then show the current icon. Then show the dropdown or inline picker without input.
3. Make sure to indicate "no selection" if there is no icon specified in the metadata file, while showing the gear/cog icon which is default.
4. Supporting 4.0 aliases means: if a valid alias is currently in use, continue to use that alias. Maybe if it is a recognized alias we should show the corresponding FA5 icon selected in the dropdown, but a separate checkbox saying "[*] Use Font Awesome 4 compatible aliases". So if `fa://gear` is selected, and you select another icon and then select cog again, the FA4 alias "gear" is used. If you uncheck the "Use Font Awesome 4 compatible aliases" then "cog" would be used. The  "Use Font Awesome 4 compatible aliases" checkbox can be present in the form at all times, but by default unchecked - unless the metadata file previously had a FA4 alias in use. This means that if the metadata file has `fa://gear` and I open the edit medatata file to change the title, the title should change, but the `fa://gear` should stay the same after I save.

laser505t:
Fixes #380

### DIFF
--- a/app/models/dev_product.rb
+++ b/app/models/dev_product.rb
@@ -25,5 +25,4 @@ class DevProduct < Product
   def permissions?
     false
   end
-
 end


app/views/products/_form_git.html.erb
```diff
@@ -1,5 +1,5 @@
 <div class="field">
-  <%= f.text_field :git_remote, help: "Example: <strong>git@github.com:Example/example.git</strong> (<a href='https://help.github.com/articles/generating-an-ssh-key/' target='_blank'>requires SSH key installed</a>) or <strong>https://github.com/Example/example.git<strong>".html_safe %>
+  <%= f.text_field :git_remote, help: "Example: <strong>git@github.com:Example/example.git</strong> (<a href='https://help.github.com/articles/generating-an-ssh-key/' target='_blank'>requires SSH key installed</a>) or <strong>https://github.com/Example/example.git</strong>".html_safe %>
```
MorganRodgers: 👍

PR #384, opened 2018-08-09
WIP: Remove app delete feature

### LINKED ISSUE #234: Get rid of the "trash" feature and deleting an app
Instead, deleting an app should show you a page with directions on how to delete it, and a link to the shell, i.e. open the shell to this directory, then type rm -rf app

laser505t:
Fixes #234

### DIFF
--- a/app/controllers/products_controller.rb
+++ b/app/controllers/products_controller.rb
@@ -77,11 +77,19 @@ class ProductsController < ApplicationController
   # DELETE /products/1.json
   def destroy
     @type = params[:type].to_sym
-    @product = Product.find(@type, params[:name])
 
-    @product.destroy
+    begin
+      @product = Product.find(@type, params[:name])
+    rescue Product::NotFound
+      respond_to do |format|
+        format.html { redirect_to products_url(type: @type), notice: 'The app was successfully deleted.' }
+        format.json { head :no_content }
+      end
+      return
+    end
+
     respond_to do |format|
-      format.html { redirect_to products_url(type: @type), notice: 'Product was successfully moved to the trash.' }
+      format.html { redirect_to products_url(type: @type), alert: "Oh no! The app does not appear to have been removed. Either you did not issue the correct command, or a file may have had open handles with one or more processes." }
       format.json { head :no_content }
     end
   end
--- a/app/models/product.rb
+++ b/app/models/product.rb
@@ -93,14 +93,6 @@ class Product
     rescue Errno::EACCES
       false # user does not have permission to create this directory
     end
-
-    def trash_path
-      router.base_path.join(".trash")
-    end
-
-    def trash_contents
-      trash_path.directory? ? trash_path.children : []
-    end
   end
 
   def app
@@ -171,11 +163,6 @@ class Product
     else
       false
     end
-  end
-
-  def destroy
-    self.class.trash_path.mkpath
-    FileUtils.mv router.path, self.class.trash_path.join("#{Time.now.localtime.strftime('%Y%m%dT%H%M%S')}_#{name}")
   end
 
   def permissions?
--- a/app/views/products/index.html.erb
+++ b/app/views/products/index.html.erb
@@ -7,13 +7,6 @@
   <span class="pull-right">
     <%= link_to 'Launch Shell', OodAppkit.shell.url(host: ENV['OOD_DEV_SSH_HOST'], path: product_class.router.base_path.realdirpath).to_s, target: '_blank', class: 'btn btn-default' %>
     <%= link_to 'Launch Files', OodAppkit.files.url(path: product_class.router.base_path.realdirpath).to_s, target: '_blank', class: 'btn btn-default' %>
-    <% unless product_class.trash_contents.empty? %>
-      <%= link_to "Open Trash <span class=\"badge\">#{product_class.trash_contents.size}</span>".html_safe,
-          OodAppkit.files.url(path: product_class.trash_path.realpath).to_s,
-          target: '_blank',
-          class: 'btn btn-default'
-      %>
-    <% end %>
   </span>
 </p>
 
--- a/app/views/products/show.html.erb
+++ b/app/views/products/show.html.erb
@@ -155,11 +155,13 @@
   method: :delete,
   data: {
     original_title: 'Are you ABSOLUTELY sure?',
-    confirm: '<div class="alert alert-warning">Unexpected bad things will happen if you don\'t read this!</div><p>This will move the product to the trash. Any users currently running this app will experience breaking behavior. Any past or present users of this app will be unable to access their data for this app. Be sure you know what you are doing!</p>'.html_safe,
-    verify: @product.name,
-    verify_text: 'Please type in the directory name of the app to confirm.',
-    commit: 'I understand the consequences, delete this app',
-    commit_class: 'btn-danger btn-block',
+    confirm: "<div class=\"alert alert-warning\">Unexpected bad things will happen if you don\'t read this!</div>
+              <p>This will permanently delete this app. Any users currently running this app will experience breaking behavior. Any past or present users of this app will be unable to access their data for this app. Be sure you know what you are doing!</p>
+              <p>To delete the app, run the following command in the shell:</p>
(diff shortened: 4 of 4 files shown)


laser505t:
One small problem is that the `mv` command actually won't move a file if a process has a handle on it, even with a `-f` flag. Same with `rm -rf`. FileUtils.mv doesn't have this problem though, which was used before to send the app to the trash, and it looks like it changes the process' handle to where the file was moved to.
This will show the correct banner telling whether the app has been deleted successfully or not, most of the time. If you do it super quickly it will falsely say that it was unsuccessfully deleted. From a little testing I've done it might actually be due to delays in our network filesystem.
Pasting into the shell app with Firefox doesn't work :(

ericfranz:
So I was actually thinking we get rid of the destroy action, including the route and the controller code. The javascript button click would not open a dialog with cancel or delete buttons or be a link to delete app with method `delete`. The app would not be handling the deletion of apps. Clicking the button could show a popover instead telling the user what to do. Since deleting the app doing rm -rf can introduce problems with .nfs files preventing the deleting of the directory, instead I think you can move it to tmp instead and then rm -rf that tmpdir:

```
mktemp -d | xargs -I % sh -c 'echo %; mv app_dir %; rm -rf %;'
```

This could be in a popover for the delete button (that does nothing else but shows the popover):


* in this screenshot "bin" is the name of the app dir - sorry, bad choice but you get the picture

Or just have a red panel or alert box at the bottom explaining what to do:



Of course, there may be a simpler recommended way to delete this directory from the command line that Trey or someone else recommends. But thats the idea - just plain Bootstrap HTML. Of course you could have "terminal" link to the terminal, in the parent directory of the app dir.

ericfranz:
@treydock when a developer goes to delete an app from the terminal, the app may have already been launched (and thus processes have those files open). So if I do `rm -rf app_dir` the removal fails and .nfs00000203003 files appear in the directory. However, I can move the directory safely.

So is there a problem with this command instead (or something similar) - move the directory to tmpdir, then rm -rf the directory. Any empty directory and .nfs00000 files remaining can just be ignored, and tmp will be periodically cleaned up.

```
mktemp -d | xargs -I % sh -c 'echo %; mv app_dir %; rm -rf %;'
```

Or is there a preferable way to remove this without worrying about killing the web processes with these files open?

ericfranz:
@laser505t if you go the panel route, might look better aligned with the README etc. at bottom:



Though that screenshot it looks kinda silly drawing all that attention to something users probably would rarely do.

treydock:
@ericfranz If not killing off processes then your example of moving files to /tmp is probably best option.

laser505t:
I don't know if this will work completely. Like I said before, I think Ruby's FileUtils.mv is rather smart compared to the binary `mv` command:
```
-bash-4.2$ ls bar
file2.txt  file.txt
-bash-4.2$ cat > bar/file.txt &
[1] 161065
-bash-4.2$ jobs
[1]+  Stopped                 cat > bar/file.txt
-bash-4.2$ mktemp -d | xargs -I % sh -c 'echo %; mv bar %; rm -rf %;'
/tmp/tmp.iTDkmi8FDv
mv: cannot remove ‘bar’: Directory not empty
-bash-4.2$ ls bar
-bash-4.2$ ls -a bar
.  ..  .nfs000000000004de550000130a
-bash-4.2$
```

ericfranz:
No its not smarter. The error is due to the fact that we are moving the directory from one file system to another. When the mv occurs on the same file system, the directory inode number stays the same when it is moved to another directory on the same device.

```
efranz@webtest04:~/tmp/mvtest$ ls -id bar1
151666 bar1
efranz@webtest04:~/tmp/mvtest$ mv bar1 bar10
efranz@webtest04:~/tmp/mvtest$ ls -id bar10
151666 bar10
```

That is why the original "mv to trash" worked while my "move to tmp then delete" did not work: 

```ruby
irb(main):002:0> FileUtils.mv 'bar2', '/tmp/bar2'
Errno::ENOTEMPTY: Directory not empty @ dir_s_rmdir - bar2
	from /opt/rh/rh-ruby22/root/usr/share/ruby/fileutils.rb:1445:in `rmdir'
	from /opt/rh/rh-ruby22/root/usr/share/ruby/fileutils.rb:1445:in `block in remove_dir1'
	from /opt/rh/rh-ruby22/root/usr/share/ruby/fileutils.rb:1456:in `platform_support'
	from /opt/rh/rh-ruby22/root/usr/share/ruby/fileutils.rb:1444:in `remove_dir1'
	from /opt/rh/rh-ruby22/root/usr/share/ruby/fileutils.rb:1437:in `remove'
	from /opt/rh/rh-ruby22/root/usr/share/ruby/fileutils.rb:779:in `block in remove_entry'
	from /opt/rh/rh-ruby22/root/usr/share/ruby/fileutils.rb:1494:in `ensure in postorder_traverse'
	from /opt/rh/rh-ruby22/root/usr/share/ruby/fileutils.rb:1494:in `postorder_traverse'
	from /opt/rh/rh-ruby22/root/usr/share/ruby/fileutils.rb:777:in `remove_entry'
	from /opt/rh/rh-ruby22/root/usr/share/ruby/fileutils.rb:534:in `rescue in block in mv'
	from /opt/rh/rh-ruby22/root/usr/share/ruby/fileutils.rb:527:in `block in mv'
	from /opt/rh/rh-ruby22/root/usr/share/ruby/fileutils.rb:1571:in `block in fu_each_src_dest'
	from /opt/rh/rh-ruby22/root/usr/share/ruby/fileutils.rb:1585:in `fu_each_src_dest0'
	from /opt/rh/rh-ruby22/root/usr/share/ruby/fileutils.rb:1569:in `fu_each_src_dest'
	from /opt/rh/rh-ruby22/root/usr/share/ruby/fileutils.rb:517:in `mv'
	from (irb):2
	from /opt/rh/rh-ruby22/root/usr/bin/irb:11:in `<main>'
irb(main):003:0> FileUtils.mv 'bar2', 'bar4'
=> 0
```

So my suggestion is bad. Probably better to move the directory to a tmp location on the same file system. In the user's home directory that would be ~/tmp/. Obviously that is by convention. Move there, then recursive delete would throw an error and a .nfs00000 file would remain but then when it was gone an empty directory would remain.

ericfranz:
So yeah this would work:

```
efranz@webtest04:~/tmp/mvtest$ mktemp -d --tmpdir=$HOME/tmp/ | xargs -I % sh -c 'echo %; mv bar4 %; rm -rf %;'
/users/PZS0562/efranz/tmp/tmp.0CFFOvSyL3
rm: cannot remove `/users/PZS0562/efranz/tmp/tmp.0CFFOvSyL3/bar4/.nfs00000000000250750000000e': Device or resource busy
efranz@webtest04:~/tmp/mvtest$ ls
.  ..  bar10
efranz@webtest04:~/tmp/mvtest$ mktemp -d --tmpdir=$HOME/tmp/ | xargs -I % sh -c 'echo %; mv bar10 %; rm -rf %;'
/users/PZS0562/efranz/tmp/tmp.57zdZ3enc2
rm: cannot remove `/users/PZS0562/efranz/tmp/tmp.57zdZ3enc2/bar10/.nfs00000000000250740000000d': Device or resource busy
efranz@webtest04:~/tmp/mvtest$ ls
.  ..
```

Well make sure ~/tmp exists:

```
mkdir -p $HOME/tmp; mktemp -d --tmpdir=$HOME/tmp/ | xargs -I % sh -c 'echo %; mv bar10 %; rm -rf %;'
```

Thats a little ugly looking. But it is only a suggestion to the user. The directions could be:

>Remove this app using the terminal. If any processes have files open in this directory, you may run into problems. Then you can move the directory to a tmp dir on the same file system. For example, I move it to my home directory and then delete it:
>
>     mkdir -p $HOME/tmp; mktemp -d --tmpdir=$HOME/tmp/ | xargs -I % sh -c 'echo %; mv bar10 %; rm -rf %;'
>
> The command may throw an error, but if you inspect the directory only a .nfs0000... file may remain. After the processes using this directory exit, the empty directory would remain in tmp.

ericfranz:
We did https://github.com/OSC/ood-dashboard/pull/394 instead

PR #385, opened 2018-08-13
add whitelist mode and set to false by default

### LINKED ISSUE #295: Remove whitelist of apps that appear in navbar by default
There should be a "whitelist" mode. When it is false, the "whitelist" array (currently NavConfig.categories simply specifies the order of the first categories that appear. Any other categories are auto-appended to the end of this list. Then we would default it to false and instruct users who are updating (OOD 1.4?) to enable whitelist mode if suddenly apps appear.

#### Old description:

>Instead of having a default whitelist for the navbar be an ordered list of categories:
>
>- Files, Jobs, Clusters, Interactive Apps
>
>Lets remove the whitelist by default. If we did that we would order alphabetically, resulting in:
>
>- Clusters, Files, Interactive Apps, Jobs
>
>Since Files are used the most, we would rename Clusters to Systems, and the default OOD installation would result in this navigation:
>
>- Files, Interactive Apps, Jobs, Systems
>
>This way, if I add a new app to my default OOD install with the category "Reports" it will immediately appear in the navbar, without any modification in an initializer to specify the whitelist or change the order:
>
>- Files, Interactive Apps, Jobs, Reports, Systems
>
>
>For OSC, we could still use our whitelist to limit what drop downs appear (so if we just want Files, Interactive Apps, Jobs, Systems in the top menu, and All Apps to access other apps in other categories in the future).
>
>This seems like a small change that could make custom app deployment easier at other centers.
>

AriettaZ:
### Default navconfig:
```
  self.categories = ["Files", "Jobs", "Clusters", "Interactive Apps"]
  self.categories_whitelist = false
```

### Views:

- NavConfig.categories is empty and whitelist is enabled results in no app menus being created

- NavConfig.categories is set to default and whitelist is enabled results in app menus excluding gateways app

- NavConfig.categories is set to default listing and whitelist is disabled results in app menus including gateways app menu in end

- NavConfig.categories is set to nil or empty array and whitelist is disabled means that all the possible app menus appear, but in alphabetical order

 

### Issues:
1. osc-editor and ood-dashboard are in the System Installed Apps category, so they will be after gateway apps.  Do we want them to stay in the navbar?
2. Do I need to modify categories list according to old description of issue 295?
>
>  Instead of having a default whitelist for the navbar be an ordered list of categories:
> 
> Files, Jobs, Clusters, Interactive Apps
> Lets remove the whitelist by default. If we did that we would order alphabetically, resulting in:
> 
> Clusters, Files, Interactive Apps, Jobs
> Since Files are used the most, we would rename Clusters to Systems, and the default OOD installation would result in this navigation:
> 
> Files, Interactive Apps, Jobs, Systems
> This way, if I add a new app to my default OOD install with the category "Reports" it will immediately appear in the navbar, without any modification in an initializer to specify the whitelist or change the order:
> 
> Files, Interactive Apps, Jobs, Reports, Systems
> For OSC, we could still use our whitelist to limit what drop downs appear (so if we just want Files, Interactive Apps, Jobs, Systems in the top menu, and All Apps to access other apps in other categories in the future).
> 
> This seems like a small change that could make custom app deployment easier at other centers.

fix #295

### DIFF
--- a/app/apps/nav_config.rb
+++ b/app/apps/nav_config.rb
@@ -3,7 +3,7 @@ class NavConfig
     attr_accessor :categories, :categories_whitelist
   end
   self.categories = ["Files", "Jobs", "Clusters", "Interactive Apps"]
-  self.categories_whitelist = true
+  self.categories_whitelist = false
   def self.categories_whitelist?
       self.categories_whitelist
   end
--- a/app/apps/nav_config.rb
+++ b/app/apps/nav_config.rb
@@ -1,6 +1,10 @@
 class NavConfig
   class << self
-    attr_accessor :categories
+    attr_accessor :categories, :categories_whitelist
   end
   self.categories = ["Files", "Jobs", "Clusters", "Interactive Apps"]
+  self.categories_whitelist = true
+  def self.categories_whitelist?
+      self.categories_whitelist
+  end
 end
--- a/app/apps/nav_config.rb
+++ b/app/apps/nav_config.rb
@@ -1,10 +1,8 @@
 class NavConfig
-  class << self
-    attr_accessor :categories, :categories_whitelist
-  end
-  self.categories = ["Files", "Jobs", "Clusters", "Interactive Apps"]
-  self.categories_whitelist = false
-  def self.categories_whitelist?
-      self.categories_whitelist
-  end
+    class << self
+      attr_accessor :categories, :categories_whitelist
+      alias_method :categories_whitelist?, :categories_whitelist
+    end
+    self.categories = ["Files", "Jobs", "Clusters", "Interactive Apps"]
+    self.categories_whitelist = false
 end
--- a/app/apps/ood_app_group.rb
+++ b/app/apps/ood_app_group.rb
@@ -34,15 +34,11 @@ class OodAppGroup
     # only one group per title; this just makes it easy to get the group
     # Hash[ [:title1,:group1], [:title2,:group2]] => {title1: :group1, title2: :group2 }
     h = Hash[groups.map {|g| [g.title, g]}]
-
     titles.map { |t| h.has_key?(t) ? h[t] : nil }.compact
   end
   
   def self.order(titles:[], groups:[])
       h = Hash[groups.map {|g| [g.title, g]}]
-      unselected =[]
-      selected = []
-      titles.each { |t| h.has_key?(t) ? selected.push(h[t]) : unselected.push(h[t]) }
-      selected.concat(unselected.sort)
+      titles.concat(h.keys.sort).uniq.map { |t| h.has_key?(t) ? h[t] : nil }.compact 
   end
 end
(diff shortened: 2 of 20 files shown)


app/apps/nav_config.rb
```diff
@@ -1,6 +1,10 @@
 class NavConfig
   class << self
-    attr_accessor :categories
+    attr_accessor :categories, :categories_whitelist
   end
   self.categories = ["Files", "Jobs", "Clusters", "Interactive Apps"]
+  self.categories_whitelist = false
+  def self.categories_whitelist?
+      self.categories_whitelist
+  end
```
ericfranz: You can instead use alias_method. I think this works:

```ruby
class << self
  attr_accessor :categories, :categories_whitelist
  alias_method :categories_whitelist?, :categories_whitelist
end
```

test/controllers/dashboard_controller_test.rb
```diff
@@ -133,6 +137,63 @@ def dropdown_list_items(list)
...
+    
+    get :index
+    assert_response :success
+    assert_select ".navbar-collapse > .nav li.dropdown[title]", 6
+    assert_select  dropdown_link(1), text: "Files"
+    assert_select  dropdown_link(2), text: "Jobs"
+    assert_select  dropdown_link(3), text: "Clusters"
+    assert_select  dropdown_link(4), text: "Interactive Apps"
+    assert_select  dropdown_link(5), text: "Gateway Apps"
+    assert_select  dropdown_link(6), text: "System Installed Apps"
```
ericfranz: Whenever you stub class methods I think you need to unstub them at the end or else their stubbing will remain in future tests.

ericfranz: Maybe. Actually could you research that? It looks like I don't unstub in some other tests.

app/apps/ood_app_group.rb
```diff
@@ -34,7 +34,17 @@ def self.select(titles:[], groups:[])
     # only one group per title; this just makes it easy to get the group
     # Hash[ [:title1,:group1], [:title2,:group2]] => {title1: :group1, title2: :group2 }
     h = Hash[groups.map {|g| [g.title, g]}]
-
     titles.map { |t| h.has_key?(t) ? h[t] : nil }.compact
   end
+  
+  # select a subset of groups by the specified array of titles
+  # maintaining the order specified by the titles array
```
ericfranz: Fix the comment here since we aren't filtering out any groups, so "select a subset of groups by the specified array of titles" is not true.

test/controllers/dashboard_controller_test.rb
AriettaZ: I didn't find the answer from documentation about stubbing life cycle. But I created a new test method and it seems stubbed navConfig variables from the previous test methods are gone. I think the stubbed value in the test method won't carry over to future tests. However, should I still call unstub at end of the test methods just in case?

AriettaZ:
#### Updates:
1.  NavConfig.categories_whitelist=true added to initializer https://github.com/OSC/osc-ood-config/pull/15

2. CHANGELOG.md updated

Issue #376, opened 2018-07-26
Support html options for batch connect plugin attributes


ericfranz:
This code is used to create each field in the form from the given attribute definition in the form.yml.erb file:

https://github.com/OSC/ood-dashboard/blob/b2ecdd5dffdb8334a91839aa3216be1b91dda5f5/app/helpers/batch_connect/session_contexts_helper.rb#L2-L31

Notice that when calling the Rails helpers, we are restricted to options such as label, help, required, and for select fields options, for number_field min, max and step. If we expanded to support all the options that these methods handle, then doing things like adding data attributes to fields or options, adding onchange, onselect, onclick with inline JavaScript would enable adding some dynamic form functionality such as when selecting a partition, the account field is updated, or when making the selection on one dropdown, the options in another dropdown changes or is filtered, etc.

ericfranz:
Related to this, `text_area` supports options like size, rows, cols, disabled, escape, along with regular HTML attrs but these currently are not supported here.

ericfranz:
All of the fields generally follows format where options is a single hash.

Exceptions are:

1. check_box
2. select
3. resolution_field (our custom field)

Unfortunately for `select` we use `form.select attrib.id, attrib.opts.fetch(:options, []), **kwargs` to get the "choices" parameter, supply nothing for the "options" parameter or the "html_options" parameter.

Also, check_box works different, expecting something like:

```ruby
f.check_box :public, {}, "true", "false"
```

Suggested implementation:

for normal fields, we use `attrib.opts.reject { |k,v| %i(label help required).include?(k) }` as the value for "kwargs". Then we can get rid of these lines: 

```ruby


kwargs[:pattern] = attrib.opts.fetch(:pattern, nil)

**Comment by ericfranz** [comment]
Previous comment I was mistaken. Here is a clarification.

- label, help, required are options that https://github.com/bootstrap-ruby/bootstrap_form provides
- in most form fields, the field options and html options are merged in a single hash
- in select form field, the argument for the option tags is called "choices", which we unfortunately call "options" in the yaml; and then the select field options and html options are separated into two separate hashes as arguments

An example of the type of keys that appear in an attributes section include:

```
widget # our convention
label # bootstrap_form_for option
help # bootstrap_form_for option
required # bootstrap_form_for option
value # default option
min # custom form field option
max # custom form field option
step # custom form field option
options # choices hash or array for select
```

1. The field options are thus keys that are not `widget`, `options`, `html_options
[...]

PR #389, opened 2018-08-16
Improve bc template rendering

### LINKED ISSUE #369: Fix batch connect rendering problems and limitations
1. The form.yml.erb renders with binding null instead of the binding being the app
2. The submit.yml.erb renders with the binding being the SessionContext with no access to the app
3. The template renders with two objects specified "session" and "context" but no "app"
4. The "SmartAttribute" class coerces the values of all of the "opts" hash to strings, and the SessionContext uses an unclear use of `method_missing` and `respond_to_missing`, and `attributes` and `attributes=` to turn an array of "SmartAttribute" objects (set in `@attributes`) into a set of methods on the context with getters and setters, where the getters return the values coerced into a string
5. When rendering erb, if filename were specified to be the path to the erb file, then the erb code can use `__FILE__` to determine where its located (can't happen now because filename is not set)

This means that:

1. File uploads are impossible (because of the string coercion problem)
2. Doing anything with paths relative to the batch connect app code is impossible in form.yml.erb and session.yml.erb and difficult in the template (in template you have to do `BatchConnect::App.from_token(session.token).root` instead of something like `app.root` or `session.app.root`

ericfranz:
Fixes #369 and #376 

### For #369:

Just making `__FILE__` work properly in form.yml.erb and submit.yml.erb is probably MVP for most of this issue.
The file upload coercion issue I had to manually test. 

Take any interactive app, add 

```yml
attributes:
  input_file:
    widget: file_field
    label: Input File
form:
  - input_file
```

and then to submit.yml.erb do this:

```
<% raise "Format of input file: #{input_file.class} and #{input_file.inspect}" %>
```

and you will see that with this change input_file is no longer a String but the file upload object.

### For #376:

`form.yml.erb` can now accept all the field options and html options
that each form field helper can accept, including things like

```
onchange: 'alert("hi")'
```

in the select field. All field options are just key value pairs added to the hash of an attribute. All html options are added as a hash to key html_options for an attribute.

### DIFF
--- a/app/helpers/batch_connect/session_contexts_helper.rb
+++ b/app/helpers/batch_connect/session_contexts_helper.rb
@@ -4,29 +4,30 @@ module BatchConnect::SessionContextsHelper
 
     widget = attrib.widget
 
-    kwargs = {
+    field_options = attrib.opts.reject { |k,v|
+      %i(widget options html_options checked_value unchecked_value).include?(k)
+    }.merge({
       label: attrib.label(fmt: format),
       help:  OodAppkit.markdown.render(attrib.help(fmt: format)).html_safe,
       required: attrib.required
-    }
+    })
+    html_options = attrib.opts.fetch(:html_options, {})
+    all_options = field_options.merge(html_options)
+
+    checked_value = attrib.opts.fetch(:checked_value, "1")
+    unchecked_value = attrib.opts.fetch(:unchecked_value, "0")
 
     case widget
     when "select"
-      form.select attrib.id, attrib.opts.fetch(:options, []), **kwargs
-    when "number_field"
-      kwargs[:min]  = attrib.opts.fetch(:min, nil)
-      kwargs[:max]  = attrib.opts.fetch(:max, nil)
-      kwargs[:step] = attrib.opts.fetch(:step, nil)
-      form.number_field attrib.id, **kwargs
+      form.select attrib.id, attrib.opts.fetch(:options, []), field_options, html_options
     when "resolution_field"
-      resolution_field(form, attrib.id, **kwargs)
+      resolution_field(form, attrib.id, all_options)
     when "check_box"
-      form.form_group attrib.id, help: kwargs[:help] do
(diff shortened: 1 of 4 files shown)


app/models/batch_connect/app.rb
```diff
@@ -220,7 +220,13 @@ def submit_file(root:, paths: %w(submit.yml.erb submit.yml))
       # Parse an ERB and Yaml file
       def read_yaml_erb(path:, binding: nil)
         contents = path.read
-        contents = ERB.new(contents, nil, "-").result(binding) if path.extname == ".erb"
+
+        if path.extname == ".erb"
+          erb = ERB.new(contents, nil, "-")
+          erb.filename = path.to_s
+          contents = erb.result(binding)
+        end
```
ericfranz: I could probably add another method:

```ruby
def render_erb_file(path:, contents:, binding:)
  erb = ERB.new(contents, nil, "-")
  erb.filename = path.to_s
  erb.result(binding)
end
```

which could be useful/used in several places. By passing in path and contents as arguments (instead of having this new method read the contents from the path) this becomes a pure function.

ericfranz: FWIW there is a one-liner that would allow this:

```ruby
ERB.new(contents, nil, "-").tap {|erb| erb.filename = path.to_s }. result(binding)
```

But that actually reduces readability over the multiline version.

app/helpers/batch_connect/session_contexts_helper.rb
```diff
@@ -3,30 +3,20 @@ def create_widget(form, attrib, format: nil)
     return "" if attrib.fixed?
 
     widget = attrib.widget
-
```
MorganRodgers: I like that you are allowing the form methods to handle all of their own arguments.

app/models/batch_connect/app.rb
MorganRodgers: Agreed that the one liner is less readable in this situation.

PR #394, opened 2018-08-24
Remove app deletion feature

laser505t:
Replaces #384 

image

### DIFF
--- a/app/views/products/show.html.erb
+++ b/app/views/products/show.html.erb
@@ -166,9 +166,11 @@
   data: {
     toggle: 'popover',
     placement: 'auto',
-    title: 'To delete this app, run this command in the terminal',
+    title: 'Delete this app using the terminal.',
     html: 'true',
-    content: "<pre>mkdir -p $HOME/tmp; mktemp -d --tmpdir=$HOME/tmp/ | xargs -I % sh -c 'echo %; mv #{@product.app.path} %; rm -rf %;'</pre>"
+    content: "The following command moves this app to a local temporary directory and then deletes it:
+    <pre>mkdir -p $HOME/tmp; mktemp -d --tmpdir=$HOME/tmp/ | xargs -I % sh -c 'echo %; mv #{@product.app.path} %; rm -rf %;'</pre>
+    If it shows an error, then there may have been users running the app, causing open file handles to remain open and preventing the files from being completely deleted. But once these file handles close, the empty directory alone will remain."
   }
 %>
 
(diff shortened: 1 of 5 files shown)


laser505t:
Not the greatest results if you delete the folder you are in, but I'd say it's tolerable:

```
-bash-4.2$ cd bar
-bash-4.2$ ls
file2.txt  file.txt
-bash-4.2$ mkdir -p $HOME/tmp; mktemp -d --tmpdir=$HOME/tmp/ | xargs -I % sh -c 'echo %; mv /users/PZS0002/tharris/ondemand/dev/bar %; rm -rf %;'
/users/PZS0002/tharris/tmp/tmp.BkrpteFC5D
-bash-4.2$ ls
ls: cannot open directory .: Stale file handle
-bash-4.2$ pwd
/users/PZS0002/tharris/ondemand/dev/bar
-bash-4.2$ cd ..
-bash: cd: ..: No such file or directory
-bash-4.2$ cd ~
-bash-4.2$
```

app/views/products/show.html.erb
```diff
@@ -161,6 +161,12 @@
 <% end %>
 
 <%= link_to 'Back', products_path(type: @type), class: 'btn btn-default' %>
+
+<style>
+  .popover {
+    max-width: 100%;
+  }
+</style>
```
ericfranz: I think I would prefer to see `.popover-wide .popover { }`, something that can restrict it to the subset of the page that deals with this popover. When we add popovers to the dropdown menus this will affect those but we won't catch it till we navigate to this page.

ericfranz: I confirmed, for example, if you specify an explicit id on a div and in the popover construction set the data attribute "container" to the id string the popover html will be created as a child of that container.

ericfranz:
Sorry, it just occurred to me after, that gee thats a really long command, usually when we have that we break it up into multiple lines so its easier to read. Something like:

```shell
APP_DIR=/users/PZS0002/tharris/ondemand/dev/bar && \
mkdir -p $HOME/tmp && \
mktemp -d --tmpdir=$HOME/tmp/ | \
xargs -I % sh -c "echo % && mv $APP_DIR % && rm -rf %;"
```

And then I realized, if we do this, the width of the popover is smaller. And maybe the custom container/css is not necessary. But if it still is, keep it.

laser505t:
Here's just a regular popover:
image

Here's with custom width:
image

Frankly, I like neither. I don't know why the non-preformatted text is dictating the width the way it is, but the default seems too cramped.

ericfranz:
Here we go. Popover is the wrong control: https://getbootstrap.com/docs/3.3/javascript/#collapse-example

See the "Button with data-target" example. Just put the HTML for deleting an app in a panel underneath the Delete button.

laser505t:
image
(after clicking the button of course)

And it resizes very nicely!

ericfranz:
Great work!

PR #410, opened 2018-10-10
Update hardcoded scl references

### LINKED ISSUE #408: Don't hardcode SCLs in dashboard
Instead, use something like this:

```ruby
def command_scl_prefix
  if ENV['X_SCLS']
    "scl enable #{ENV['X_SCLS']} -- "
  else
    ""
  end
end
```

For display names: https://github.com/OSC/ood-dashboard/blob/ccef91024973d47613f3dfbb6a097bfb79267bc4/app/views/products/show.html.erb#L73-L76

ericfranz:
Fixes #408

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ use a database.
 2. Install the app for a production environment:
 
     ```sh
-    RAILS_ENV=production scl enable git19 rh-ruby22 nodejs010 -- bin/setup
+    RAILS_ENV=production scl enable rh-git29 rh-nodejs6 rh-ruby24 -- bin/setup
     ```
 
     this will setup a default Open OnDemand install.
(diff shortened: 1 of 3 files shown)


ericfranz:
Note: the README update is not ideal, but fastest. And while it will be incorrect for existing users going to the repo, most everyone now installs via rpm instead of manually.

ericfranz:
BTW this is what the change does:

The env you have to load based on the SCLs loaded in the NGINX config differs between installs that uses different SCLs: 

screen 2018-10-11 at 10 36 03 am

screen 2018-10-11 at 10 37 57 am

#### Which is better than hardcoding SCLs

screen 2018-10-11 at 10 39 52 am

is better than

screen 2018-10-11 at 10 39 26 am

PR #411, opened 2018-10-15
Remove implicit route to undefined controller method

laser505t:
Some changes that should have gone with #394

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -32,10 +32,10 @@ Rails.application.routes.draw do
 
   # App administration
   scope 'admin/:type', constraints: Authz::AppDeveloperConstraint do
-    resources :products, param: :name, constraints: { type: /dev|usr/ } do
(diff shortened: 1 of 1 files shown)


config/routes.rb
```diff
@@ -32,10 +32,10 @@
 
   # App administration
   scope 'admin/:type', constraints: Authz::AppDeveloperConstraint do
-    resources :products, param: :name, constraints: { type: /dev|usr/ } do
+    resources :products, except: :destroy, param: :name, constraints: { type: /dev|usr/ } do
       nested do
         scope ':context' do
-          resources :permissions, only: [:index, :new, :create, :destroy], param: :name
+          resources :permissions, only: [:index, :new, :create], param: :name
```
ericfranz: Why remove route for deleting permissions?

laser505t: Woops

PR #424, opened 2018-12-12
iHPC apps to submit to multiple clusters

ericfranz:
Client side:

1. if there is only one depdendent cluster the user can submit jobs to, that adapter is used for the "format" for client side rendering (i.e. 'format' for the bc_* attributes like bc_num_slots)
2. if there are multiple clusters of same adapter, that is used for "format" for client side rendering
3. if there are multiple clusters of different adapters, the first in the list is used (BUG - not currently fixed nor warned - also rarer edge case)

Server side:

1. server side if there is only one depdendent cluster the user can submit jobs to, that adapter is used for "format" and that id is used for cluster_id
2. server side if there is multiple dependent clusters available, and the id of one is specified in the web form submission paramter name "batch_connect_session_context[cluster]" that is used for adapter "format" and cluster_id
3. server side, if there is multiple dependent clusters available, and one is not specified, or if there are no dependent clusters available, an error is thrown

### DIFF
--- a/app/controllers/batch_connect/session_contexts_controller.rb
+++ b/app/controllers/batch_connect/session_contexts_controller.rb
@@ -23,19 +23,21 @@ class BatchConnect::SessionContextsController < ApplicationController
   # POST /batch_connect/<app_token>/session_contexts.json
   def create
     set_app
-    set_render_format
     set_session_context
+    set_specified_cluster
 
     # Read in context from form parameters
-    @session_context.attributes = session_contexts_param
+    @session_context.attributes = session_contexts_params
 
     @session = BatchConnect::Session.new
     respond_to do |format|
-      if @session.save(app: @app, context: @session_context, format: @render_format)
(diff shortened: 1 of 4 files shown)


ericfranz:
@MorganRodgers this has not been thoroughly tested yet; also this omits two things:

1. doesn't address if you declare dependencies on multiple clusters of different adapter types (torque and slurm in our future case)
2. doesn't attempt to auto create a select element that let the user set the cluster

ericfranz:
@MorganRodgers this has not been well tested nor are there comprehensive tests added. I wouldn't recommend merging this right away unless you are confident of these changes (or at least that these changes don't break anything). Regardless the code is ugly - that is partially reflective of the complexity of the code that was modified.

ericfranz:
Some of @MorganRodgers thoughts from an Asana task:

- Some conditionals are incredibly long and should be extracted into methods
- Better automated tests would be great
- A Vagrant setup where we could test multiple clusters / queues would be great (quick with Slurm, slower with SGE)

PR #425, opened 2018-12-12
Update default logout wording fixes #420

### LINKED ISSUE #420: With Basic Auth, logout appearance somewhat misleading
The default logout page (at least in a Basic Auth scenario) should more emphatically state that _until_ the browser is closed, the user is still logged in (even though they've hit the logout button).

Also, I wonder if one of the techniques on [this page](https://stackoverflow.com/questions/233507/how-to-log-out-user-from-web-site-using-basic-authentication) couldn't be utilized here to log the user out without closing the browser.  In particular, the javascript bookmarklet seems to work for current Chrome and Firefox, at least.  Not exactly sure how to add that myself, though.

ericfranz:
1. make the default logout message more clear as to the actions users must take to completely logout
2. address edgecase when using Chromebook
3. suggest using private browsing mode

Fixes https://github.com/OSC/ood-dashboard/issues/420

screen 2018-12-12 at 3 01 55 pm

### DIFF
--- a/app/views/dashboard/logout.html.erb
+++ b/app/views/dashboard/logout.html.erb
@@ -1,5 +1,15 @@
 <!-- For basic authentication, we ask the user to close their browser -->
 <div class="jumbotron">
-  <h2>Log Out</h2>
-  <p class="lead">Please close your browser to log out. If after reopening your browser you are still logged in, please clear your cookies.</p>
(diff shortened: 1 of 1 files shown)


michaelkarlcoleman:
Looks good!

PR #426, opened 2018-12-12
Custom js for interactive apps

ericfranz:
Add custom javascript at bottom of interactive app form if `form.js` or `#{sub_app_name}.js` exist (i.e. `owens-vdi.js`).

### DIFF
--- a/app/models/batch_connect/app.rb
+++ b/app/models/batch_connect/app.rb
@@ -175,6 +175,26 @@ module BatchConnect
       file.read if file.file?
     end
 
+    # Path to custom javascript file for app or subapp
+    # @return [Pathname] path to custom javascript app
+    def custom_javascript_file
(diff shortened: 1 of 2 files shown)


app/models/batch_connect/app.rb
```diff
@@ -175,6 +175,14 @@ def session_view
       file.read if file.file?
     end
 
+    # Paths to custom javascript files
+    # @return [Pathname] paths to custom javascript files that exist
+    def custom_javascript_files
+      files = [root.join("form.js")]
+      files << sub_app_root.join("#{sub_app}.js")
```
MorganRodgers: Why permit loading a second script? Do you just want to give developers the option of something other than `form.js`? Or do you envision a use case where the second script is necessary?

PR #431, opened 2018-12-18
Use new dev apps env var OOD_DEV_APPS_ROOT

ericfranz:
dev apps base path is now to be defined in OOD_DEV_APPS_ROOT

nginx_stage will set this to whatever is configured in nginx_stage.yml
config

this ensures that default dev app setup in OOD 1.4 will work without requiring
extra steps like adding this code to an initializer:

    Configuration.app_development_enabled = File.directory? "/var/www/ood/apps/dev/#{OodSupport::Process.user.name}/gateway"

### DIFF
--- a/app/apps/dev_router.rb
+++ b/app/apps/dev_router.rb
@@ -26,7 +26,7 @@ class DevRouter
   end
 
   def self.base_path(owner: OodSupport::Process.user.name)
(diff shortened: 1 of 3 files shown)


ericfranz:
Note:

```
irb(main):002:0> Pathname.new("/dev/null").realpath.directory?
=> false
irb(main):003:0> Pathname.new("/dev/null").realpath
=> #<Pathname:/dev/null>
irb(main):004:0> Pathname.new("/dev/null").directory?
=> false
```

So it is useful in certain cases.

PR #432, opened 2018-12-19
Add RSS alerts to dashboard home page

laser505t:


config/configuration_singleton.rb
```diff
@@ -145,6 +145,10 @@ def dataroot
 
     Pathname.new(root).expand_path
   end
+  
+  def rss_announcements_url
+    ENV['RSS_ANNOUNCEMENTS_URL']
```
ericfranz: Prefix the env var with OOD_ i.e. OOD_RSS_ANNOUNCEMENTS_URL

ericfranz:
Move the alerts view to the layout where the other announcement alerts are displayed. That way they will appear in the same location as the other alerts.

ericfranz:
Putting this on hold for now. I think that https://github.com/OSC/ood-dashboard/pull/462#issuecomment-482273249 comment has an idea for refactoring the announcement code, and https://github.com/OSC/ood-dashboard/issues/441#issuecomment-476932884 comment suggests that we should treat this feature similar to how we treat the quota feature: you can add URLs along with PATHS to the ANNOUNCEMENT_PATH, and if it is a URL we will fetch it. However, we still want to do the URL fetch client side instead of server side.

ericfranz:
This change set is small enough we can copy and paste when we revisit. We want to support both "warning" and "info" messages, and need a way to support this. Needs more design.

PR #433, opened 2018-12-27
Add version string to footer

ericfranz:
Display VERSION file, or git describe for the version string.

### DIFF
--- a/app/views/layouts/_footer.html.erb
+++ b/app/views/layouts/_footer.html.erb
@@ -1,14 +1,20 @@
 <footer>
   <div class="container-fluid">
-    <hr>
-    <%= link_to "https://osc.github.io/Open-OnDemand/" do %>
-      <%= image_tag(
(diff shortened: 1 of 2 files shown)


ericfranz:
If this is a good solution in the short term, this could probably be added to a version class in OodAppkit so then instead of using `Configuration.version` we would use `OodAppkit.app_version` or something similar.

ericfranz:
When adding the OnDemand version we could also add the `OOD_VERSION` and `OOD_PORTAL` to nginx_stage, and deprecate `ONDEMAND_VERSION` and `ONDEMAND_PORTAL`.

app/views/layouts/_footer.html.erb
```diff
@@ -1,14 +1,20 @@
 <footer>
   <div class="container-fluid">
-    <hr>
```
MorganRodgers: This `hr` went away.

Issue #417, opened 2018-11-09
Processing erb on each request to generate the description information produces performance issues


ericfranz:
We load and process the batch connect form.yml.erb files to generate the title and description of the iHPC apps. Unfortunately, this means that any expensive parts of the form.yml.erb are also executed. If there is code to build a queue list from a command, for example, that will be executed for each iHPC app on every request to the dashboard. This is terrible for performance.

ericfranz:
A bigger problem though is the fact that the title and description can be set in the form.yml for ihpc apps, and if its a form.yml.erb, that erb could potentially affect the title and description.

Even with a solution like https://github.com/OSC/ood-dashboard/issues/418 the form.yml.erb will still be processed when building the cache, whether it is done every time a new Passenger app instance loads (which means it would be re-done every 5 minutes if a user is inactive for 5 minutes after accessing the dashboard).

We have several cases:

1. I have a form.yml. I'd prefer a cache for the menu and don't care if users have to restart their PUNs or wait for their Passenger apps to stop before seeing the yaml update.
2. I have a form.yml. I'd prefer caching but proper cache invalidation if the form.yml is modified.
3. I have a form.yml.erb that specifies a static title and description in the yaml, but contains expensive erb code for getting a list of queues, etc. when building the web form. I'm not doing app sharing or heavy app development. **I want the title and description to display without the erb being processed for iHPC apps and I want to cache the apps list at PUN start.** 
4. I have a form.yml.erb that uses a dynamic title and description. I want the title and description to update on every request.
5. I have a form.yml.erb that uses a dynamic title and description. I want the title and description to update when the menu cache is built.
6. I have form.yml.erb that uses a dynamic title and description and other erb code that is expensive. I want the title and description to update when the menu cache is built but I only want to have the expensive erb processed when the web form is accessed.

ericfranz:
This problem gets a little worse.

1. We have this concept of "subapps" in batch connect that is used solely by the bc_desktop app. So the subapp specifies a title and description in the form.yml.erb file i.e. for us at /etc/ood/config/apps/bc_desktop/ we have oakley.yml which defines `title: "Oakley Desktop"`
2. We have the idea of form validation where each form specifies a cluster (oakley, ruby) and then if you do not have access to that cluster (ruby is restricted) you do not see the app in the navigation
3. As a result of 1 & 2 above, we load the form config (which includes the erb processing), when building the navigation. This is currently done on each request.
[...]

PR #435, opened 2018-12-28
Fix displaying batch connect launch btn

ericfranz:
in the developer views, the batch connect
launch buttons would not appear if the batch connect app was
not a "valid" app - which is the case when you start developing
a new app that doesn't specify a required cluster, for example, or
if you write some code that breaks the erb rendering

this ensures that the launch buttons always display in the developer views

### DIFF
--- a/app/models/batch_connect/app.rb
+++ b/app/models/batch_connect/app.rb
@@ -102,6 +102,18 @@ module BatchConnect
       ood_app.manifest.description
     end
 
+    def link
+      OodAppLink.new(
+        # FIXME: better to use default_title and "" description
+        title: title,
+        description: description,
+        url: Rails.application.routes.url_helpers.new_batch_connect_session_context_path(token: token),
+        icon_uri: ood_app.icon_uri,
+        caption: ood_app.caption,
+        new_tab: false
+      )
+    end
+
     # Cluster id the batch connect app uses
     # @return [String, nil] cluster id used by app
     def cluster_id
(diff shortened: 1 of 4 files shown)


ericfranz:
Several problems with this PR

1. The refactor and the fix are in the same commit, instead of separate
2. The refactor introduces tighter coupling between OodApp and BatchConnect::App classes (though it was originally suggested that BatchConnect::App should have been a subclass of OodApp)
3. The "solution" punts on the original problem: that the logic of calling `select(:valid?)` on the batch connect apps is wrong to begin with. When building the menu of apps the user should launch, the only question is if an app is for a cluster the user does not have access to (i.e. Ruby Desktop). We want to avoid erb rendering during this menu building process as well.

So this may be a good short term solution - so that the app development tutorial in the documentation actually works - but we need to revisit.

app/models/batch_connect/app.rb
```diff
@@ -102,6 +102,18 @@ def default_description
       ood_app.manifest.description
     end
 
+    def link
+      OodAppLink.new(
+        # FIXME: better to use default_title and "" description
```
MorganRodgers: Would this fix take too long?

ericfranz: No. It changes functionality though.

ericfranz: I added a comment to https://github.com/OSC/ood-dashboard/issues/417 to address this.

PR #436, opened 2018-12-31
Handle invalid limit in quota

ericfranz:
Fix edge case where quota limit is invalid, so has the value 0.
In this case we treat it like a sufficient quota so we would not warn the user.

TODO:

1. Handle other "crash" edge cases, where any of the required quota params passed to `Quota.create_both_quota_types` from the parsed file are missing. In this case we should not crash, and instead log a warning to the user. Every call to "fetch" will raise an exception.
2. Ensure that for invalid quota cases we flag this as invalid and we log a warning to the logs. 
3. Try to reproduce original problem to determine why the limit was 0 in the first place. This is something we will want to search for in the logs afterwards to see why users (such as Summer and Samuel) have the experience of having 0 set for the limit (and why this is not reproduceable for users like efranz). It might be worth setting this branch up in ondemand-dev and seeing if we can reproduce the problem there with the help of Samuel or Summer.

### DIFF
--- a/app/models/quota.rb
+++ b/app/models/quota.rb
@@ -99,7 +99,11 @@ class Quota
   end
 
   def sufficient?(threshold: 0.95)
-    @total_usage < threshold * @limit
+    if valid?
+      @total_usage < threshold * @limit
+    else
+      true
+    end
   end
 
   def insufficient?(threshold: 0.95)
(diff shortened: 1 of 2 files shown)


ericfranz:
Actually I think the approach I was taking was the wrong one. Sure it protects against the divide by zero.

But for the rest of the quota format parsing it would seem like there should be validation of the original “create_both_quota_types”.

And an “InvalidJson” exception raised like the one that is already handled earlier. And this should result in a warning log to the Rails log.

MorganRodgers:
The source of this problem is that OSC doesn't have a per-user policy limiting scratch usage, so the limits are reported as 0 in the quota entries.

The solution is to provide an alternative view for non-policy-limited storage telling the user how much they are using.

Issue #442, opened 2019-01-28
Add localization


ericfranz:
1. Dashboard (immediate)
   - [x] 1. tagline = OnDemand provides an integrated, single access point for all of your HPC resources.
    - [x] 1. welcome section (replace logo, tagline, etc. with a single block of markdown and pass logo URL as the argument to this block; make h6 lead style) OR just a hook for more text
   - [x] 1. motd title = Message of the Day
   - [x] 1. quota (sounds harder, maybe wait)
1. Dashboard
   - [ ] 1. Develop dropdown
   - [x] 1. Help dropdown
   - [x] 1. Logged in as string
   - [x] 1. Log Out button
   - [x] 1. support link - If you don't see the app you are looking for, please contact support.
   - [ ] 1. nabar (category names and app names)
   - [x] 1. shared apps explanation "Cluster access applications are available from the dropdown menus in the navbar above. Custom shared apps are available below."
   - [x] 1. app catalog link - subset of the apps available can be viewed on the App Catalog. 
   - [ ] 1. interactive apps
      - [x] 1. You have no active sessions.
      - [x] 1. breadcrumb Home
      - [x] 1. breadcrumb My Interactive Sessions
      - [ ] 1. App category (repeated in navbar and side menu)
      - [x] 1. Session was successfully created. 
      - [x] 1. **other error messages?**
      - [x] 1. Please be patient as your job currently sits in queue. The wait time depends on the number of cores as well as time requested. 
1. JobComposer
   1. JoyRide steps
   1. New Job Options
   1. All the buttons
   1. Templates
      1. To create a new job, select a template to copy, fill out the form to the right, and click "Create New Job".
      1. Create New
      1. Selected Template Details
      1. many more...

Example sentence in diff langauge?

PR #447, opened 2019-02-01
WIP: Más translations

### LINKED ISSUE #439: Make hardcoded string "every 5 minutes" regarding quota reload rate configurable
We hard code this string: https://github.com/OSC/ood-dashboard/blob/a20cae91fb0f636c8a7e3fc8dfe508d9afa34d50/app/views/shared/_insufficient_quota_resource.html.erb#L7

Different sites might have different policies on how often the quota files are updated. This string should be configurable.

One simple solution could be to allow this string to be included in the strings that can be localized (see https://guides.rubyonrails.org/i18n.html). Then a site could change this to actually say (every 10 minutes) etc. instead of us adding special configuration options.

laser505t:
See #442 for priority list and progress tracking.
Fixes #439

### DIFF
--- a/app/controllers/batch_connect/session_contexts_controller.rb
+++ b/app/controllers/batch_connect/session_contexts_controller.rb
@@ -33,7 +33,7 @@ class BatchConnect::SessionContextsController < ApplicationController
     respond_to do |format|
       if @session.save(app: @app, context: @session_context, format: @render_format)
         cache_file.write(@session_context.to_json)  # save context to cache file
-        format.html { redirect_to batch_connect_sessions_url, notice: 'Session was successfully created.' }
+        format.html { redirect_to batch_connect_sessions_url, notice: t('dashboard.bc.sessions.status_blurb.create_success') }
         format.json { head :no_content }
       else
         format.html do
--- a/app/controllers/batch_connect/session_contexts_controller.rb
+++ b/app/controllers/batch_connect/session_contexts_controller.rb
@@ -33,7 +33,7 @@ class BatchConnect::SessionContextsController < ApplicationController
     respond_to do |format|
       if @session.save(app: @app, context: @session_context, format: @render_format)
         cache_file.write(@session_context.to_json)  # save context to cache file
-        format.html { redirect_to batch_connect_sessions_url, notice: t('dashboard.batch_connect.sessions.status_blurb.create_success') }
+        format.html { redirect_to batch_connect_sessions_url, notice: t('dashboard.tr.batch_connect_sessions_status_blurb_create_success') }
         format.json { head :no_content }
       else
         format.html do
--- a/app/controllers/batch_connect/session_contexts_controller.rb
+++ b/app/controllers/batch_connect/session_contexts_controller.rb
@@ -33,7 +33,7 @@ class BatchConnect::SessionContextsController < ApplicationController
     respond_to do |format|
       if @session.save(app: @app, context: @session_context, format: @render_format)
         cache_file.write(@session_context.to_json)  # save context to cache file
-        format.html { redirect_to batch_connect_sessions_url, notice: t('dashboard.bc.sessions.status_blurb.create_success') }
+        format.html { redirect_to batch_connect_sessions_url, notice: t('dashboard.batch_connect.sessions.status_blurb.create_success') }
         format.json { head :no_content }
       else
         format.html do
--- a/app/helpers/batch_connect/sessions_helper.rb
+++ b/app/helpers/batch_connect/sessions_helper.rb
@@ -50,7 +50,7 @@ module BatchConnect::SessionsHelper
 
   def created(session)
     content_tag(:p) do
-      concat content_tag(:strong, t('dashboard.batch_connect.sessions.stats.created_at'))
+      concat content_tag(:strong, t('dashboard.tr.batch_connect_sessions_stats_created_at'))
       concat " "
       concat Time.at(session.created_at).localtime.strftime("%Y-%m-%d %H:%M:%S %Z")
     end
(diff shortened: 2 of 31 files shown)


app/helpers/batch_connect/sessions_helper.rb
```diff
@@ -105,17 +105,17 @@ def id(session)
...
+      t('dashboard.bc.sessions.stats.status.queued')
     elsif session.held?
-      "Held"
+      t('dashboard.bc.sessions.stats.status.held')
     elsif session.suspended?
-      "Suspended"
+      t('dashboard.bc.sessions.stats.status.suspended')
     else
-      "Undetermined"
+      t('dashboard.bc.sessions.stats.status.undetermined')
```
ericfranz: Lets leave the status values un-localizable at this time. I would like to address this in a different way (by possibly introducing internationalization to ood_core).

app/controllers/batch_connect/session_contexts_controller.rb
```diff
@@ -33,7 +33,7 @@ def create
     respond_to do |format|
       if @session.save(app: @app, context: @session_context, format: @render_format)
         cache_file.write(@session_context.to_json)  # save context to cache file
-        format.html { redirect_to batch_connect_sessions_url, notice: 'Session was successfully created.' }
+        format.html { redirect_to batch_connect_sessions_url, notice: t('dashboard.bc.sessions.status_blurb.create_success') }
```
ericfranz: Instead of bc we should use the full batch_connect

```
dashboard.batch_connect.sessions.status_blurb.create_success
```

This is so that we follow similar conventions where we use batch connect in other locations i.e. https://osc.github.io/ood-documentation/master/installation/cluster-config-schema.html?highlight=batch_connect

config/locales/en.yml
```diff
@@ -25,3 +25,178 @@ en:
...
+      welcome_msg_tag: |
+        Cluster access applications are available from the dropdown menus in the navbar above.
+        <br>Custom shared apps are available below.
+      no_shared_apps_tag: |
+        <strong>No custom shared apps available.</strong>
+      support_msg_tag: |
+        If you don't see the app you are looking for, please
+        <a href="%{support_url}">contact support</a>.
+      catalog_msg_tag: |
+        <br>A subset of the apps available can be viewed on the %{catalog_link_tag}.
```
ericfranz: All of these allow actual HTML, so it should instead be `_html` instead of `_tag`. See https://guides.rubyonrails.org/v4.2/i18n.html#using-safe-html-translations.

config/locales/en.yml
```diff
@@ -25,3 +25,178 @@ en:
...
+      catalog_title: "App Catalog"
+    apps:
+      system_apps_title: "System Apps"
+      shared_apps_title: "Shared Apps"
+      setup_failure: |
```
ericfranz: `setup_failure_html` see https://guides.rubyonrails.org/v4.2/i18n.html#using-safe-html-translations

config/locales/en.yml
```diff
@@ -25,3 +25,178 @@ en:
...
+      no_shared_apps_tag: |
+        <strong>No custom shared apps available.</strong>
+      support_msg_tag: |
+        If you don't see the app you are looking for, please
+        <a href="%{support_url}">contact support</a>.
+      catalog_msg_tag: |
+        <br>A subset of the apps available can be viewed on the %{catalog_link_tag}.
+      catalog_title: "App Catalog"
+    apps:
+      system_apps_title: "System Apps"
```
ericfranz: This is actually used on the apps index page. Will you verify whether we ever actually link to this page? I think if not, we should consider deleting this page (and route) which means that `system_apps_title` key should also go away.

config/locales/en.yml
```diff
@@ -25,3 +25,178 @@ en:
...
+          missing_connection: |
+            This app is missing information required to establish a connection. Please 
+            contact support if you see this message.
+          queued: |
+            Please be patient as your job currently sits in queue. The wait time depends 
+            on the number of cores as well as time requested.
+          starting: |
+            Your session is currently starting... Please be patient as this process can 
+            take a few minutes.
+        native:
```
ericfranz: I would actually remove this native section in a separate isolated commit - along with the references in the views of course, so that we can revert the change later. The reason is that native vnc for Open OnDemand is not a documented or supported feature yet, and if admins are using this YAML as a reference for what text they can change, this makes up a large portion of the page for an undocumented feature.

It is important to revert just this and the corresponding views in an isolated commit, so that in the future we will have the ability to easily revert undoing that change if we do decide to support native vnc.

config/locales/en.yml
```diff
@@ -25,3 +25,178 @@ en:
       %{logo_img_tag}
       <p class="lead">OnDemand provides an integrated, single access point for all of your HPC resources.</p>
     motd_title: "Message of the Day"
+    breadcrumbs:
+      home: "Home"
+      my_sessions: "My Interactive Sessions"
+      all_apps: "All Apps"
+    restart_msg_tag: |
```
ericfranz: change to restart_msg_html (see https://guides.rubyonrails.org/v4.2/i18n.html#using-safe-html-translations)

config/locales/en.yml
```diff
@@ -25,3 +25,178 @@ en:
...
+        <strong>No custom shared apps available.</strong>
+      support_msg_tag: |
+        If you don't see the app you are looking for, please
+        <a href="%{support_url}">contact support</a>.
+      catalog_msg_tag: |
+        <br>A subset of the apps available can be viewed on the %{catalog_link_tag}.
+      catalog_title: "App Catalog"
+    apps:
+      system_apps_title: "System Apps"
+      shared_apps_title: "Shared Apps"
```
ericfranz: Remove `shared_apps_title` for now. Same reason for `apps.system_apps_title`. If we want to be able localize these particular strings I think we need to think more about how we want to do that since similar words are used in many different views.

config/locales/en.yml
```diff
@@ -25,3 +25,178 @@ en:
...
+          host: "Host:"
+          undetermined_host: "Undetermined"
+          session_id: "Session ID:"
+          status:
+            starting: "Starting"
+            running: "Running"
+            queued: "Queued"
+            held: "Held"
+            suspended: "Suspended"
+            undetermined: "Undetermined"
```
ericfranz: see the comment above about removing stats.status.starting, etc. for now - it belongs in ood_core

config/locales/en.yml
```diff
@@ -25,3 +25,178 @@ en:
...
+      all_apps: "All Apps"
+    restart_msg_tag: |
+      Your group membership has changed, which affects your access to apps. Your dashboard should restart automatically. 
+      If it doesn't please, click: <a href="%{restart_url}">restart web server</a>.
+    nav:
+      sessions: "My Interactive Sessions"
+      all_apps: "All Apps"
+      develop:
+        title: "Develop"
+        docs: "Developer Documentation"
```
ericfranz: Lets drop localizing the develop dropdown for now till we really do all of the develop dropdown (Develop, Developer Documentation, Sandbox Apps, Shared Apps, other?)

config/locales/en.yml
```diff
@@ -25,3 +25,178 @@ en:
...
+      help:
+        title: "Help"
+        change_password: "Change HPC Password"
+        support: "Contact Support"
+        docs: "Online Documentation"
+        two_factor: "Configure Two-Factor Authentication"
+      restart_server: "Restart Web Server"
+      user: "Logged in as %{username}"
+      logout: "Log Out"
+    bc:
```
ericfranz: Like mentioned in comment above, bc should be batch_connect

config/locales/en.yml
ericfranz: I think that too we should hold off on work on localizing the whole Developer interface right now.

ericfranz: Well... this work is actually done. Keep it. Deleting these comments.

ericfranz:
Left a bunch of comments. Also, we might consider adopting a convention, so that keys have extensions like: `_lbl` for label, or `_msg` for message, `_btn` for button label text? I don't know, its a consideration. Solicit @MorganRodgers for feedback on naming.

config/locales/en.yml
laser505t: Yes, the All Apps link in the navbar goes to that page which is enabled by the SHOW_ALL_APPS_LINK environment variable.

PR #452, opened 2019-02-27
Support fetching quota from a URL

ddietz89:
Some additions to read quota file from a URL.

- Support escaped colon (:) delimiter in `OOD_QUOTA_PATH` env variable so we can have a URL (https://...) and not have it sliced up. I originally changed this to a comma for easy coding, but that could break other sites that already use colons.

Example:

```
OOD_QUOTA_PATH="https\\://www.rcac.purdue.edu/ws/storagedirquota/$USER"
```
Needs colon double escaped (or the "\\:" is helpfully escaped and dropped before it gets to the ruby code).

- Support having timestamp embedded with each quota data point. If not available at the data point level, it will take the global timestamp at the top of the file as it does now. Example:

```
{  
   "version":1,
   "quotas":[  
      {  
         "path":"\/scratch\/rice\/d\/ddietz",
         "user":"ddietz",
         "total_block_usage":74490668,
         "block_limit":107374182400,
         "total_file_usage":"2418",
         "file_limit":"2000000",
         "timestamp":1551277018
      },
   "timestamp":"1551278383"
}
```

On a side note, my web service call returns ALL quotas for all systems so my splash page HTML code filters the data further based on the favorite paths configured in the file explorer.

 I just realized the "over quota" warning messages might pop for other system's scratch directories, but that's probably not a bad thing. And 99% of the time a user hits the limit here, it is on their global home or global project space anyway.

### DIFF
--- a/app/models/quota.rb
+++ b/app/models/quota.rb
@@ -1,4 +1,6 @@
 # This describes disk quota utilization for a given user and volume
+require 'net/http'
+require 'uri'
 class Quota
   class InvalidQuotaFile < StandardError; end
 
@@ -18,8 +20,35 @@ class Quota
       user  = user && user.to_s
 
       quotas = []
+
+      # Attempt to convert path into a URI
+      uri = URI.parse(quota_path)
+      if uri.instance_of?(URI::HTTP) or uri.instance_of?(URI::HTTPS)
+        # If it is a URI, and it is http:// or https://
+        begin
+          raw = Net::HTTP.get(uri)
+        rescue Exception => e
+            # There are a million ways this could go wrong, assume configured correctly and is temporary issue (e.g., not a URL typo).
+            Rails.logger.error("Quota URI failed to return data: #{e.message}")
+            # Bail with empty results. Don't break portal because web service is down.
+            return []
+        end
+      else
+        # If not a URL, assume it is a local file and attempt to read.
+        # Assume this always works, unless configured wrong, in which case don't attempt to catch.
+        raw = quota_path.read
+      end
+
+      # Attempt to parse raw JSON into an object
       begin
-        json = JSON.parse(quota_path.read)
+        json = JSON.parse(raw)
+      rescue JSON::ParserError => e
+        Rails.logger.error("Quota file is not limited JSON: #{e.message}")
+      end
+
+      Rails.logger.error("#{json}")
+      # Parse JSON object into quota data
+      begin
         case json["version"].to_i
         when 1
           quotas += find_v1(user, json)
@@ -28,8 +57,6 @@ class Quota
         end
       rescue KeyError => e
         Rails.logger.error("Quota entry for user #{user} is missing expected parameter #{e.message}")
-      rescue JSON::ParserError => e
-        Rails.logger.error("Quota file is not limited JSON: #{e.message}")
       end
 
       quotas
@@ -42,8 +69,13 @@ class Quota
       raise InvalidQuotaFile.new("Quota file with version 1 formatting missing quotas array section") unless params["quotas"].respond_to?(:each)
 
       q = []
-      time = params["timestamp"]
       params["quotas"].each do |quota|
+        # If individual quota data points include a timestamp, use that instead of the global source timestamp
+        if quota.key?("timestamp")
+          time = quota["timestamp"]
+        else
+          time = params["timestamp"]
+        end
         q += create_both_quota_types(quota.merge "updated_at" => time) if !user || user == quota["user"]
       end
       q
--- a/app/models/quota.rb
+++ b/app/models/quota.rb
@@ -22,8 +22,8 @@ class Quota
       quotas = []
 
       # Attempt to convert path into a URI
-      uri = URI.parse(quota_path)
-      if uri.instance_of?(URI::HTTP) or uri.instance_of?(URI::HTTPS)
+      if quota_path.match(/^https?:/)
+        uri = URI.parse(quota_path)
         # If it is a URI, and it is http:// or https://
         begin
           raw = Net::HTTP.get(uri)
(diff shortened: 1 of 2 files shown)


ddietz89:
Let me run the tests locally and see if I can catch the CI errors.

ddietz89:
Ok, I think it passes the CI tests now. The test was feeding `quota.rb` a Pathname, but I had changed it to take a string and figure it out then. Rather than mess with test, I changed it to accept a string (local file or URL) or a Pathname so that should be more robust.

app/models/quota.rb
```diff
@@ -18,8 +20,39 @@ def find(quota_path, user)
...
+        end
+        # Assume this always works, unless configured wrong, in which case don't attempt to catch.
+        raw = quota_path.read
+      end
+
+      # Attempt to parse raw JSON into an object
+      begin
+        json = JSON.parse(raw)
+      rescue JSON::ParserError => e
+        Rails.logger.error("Quota file is not limited JSON: #{e.message}")
```
ericfranz: If reading the json string fails, `json` will be nil, but we will attempt to use it in the case statement below, which will result in an exception.

Simplest fix is to `return []` similar to what you do above when handling exception raised by `Net::HTTP.get(uri)`

app/models/quota.rb
```diff
@@ -18,8 +20,39 @@ def find(quota_path, user)
...
 
       quotas = []
+
+      # Attempt to convert path into a URI
+      if quota_path.instance_of?(String) and quota_path.match(/^https?:/)
+        uri = URI.parse(quota_path)
+        # If it is a URI, and it is http:// or https://
+        begin
+          raw = Net::HTTP.get(uri)
+        rescue Exception => e
```
ericfranz: Its best to rescue `StandardError` instead of `Exception`. Rescuing `Exception` will rescue exceptions we don't want to rescue like `NoMemoryError`, `SyntaxError` `SignalException` etc.

app/models/quota.rb
ddietz89: Good point, will do that.

app/models/quota.rb
ddietz89: Thanks, this is helpful. I'm new to Ruby so still learning.

config/configuration_singleton.rb
```diff
@@ -80,7 +80,7 @@ def announcement_path
   # The paths to the JSON files that store the quota information
   # @return [Array<Pathname>] quota paths
   def quota_paths
-    ENV.fetch("OOD_QUOTA_PATH", "").strip.split(":").map {|p| Pathname.new(p)}
+    ENV.fetch("OOD_QUOTA_PATH", "").strip.split(/(?<!\\):/).map {|p| p.gsub(/\\/,"")}
```
ericfranz: One final thought:

The ability to escape colon in a colon delimited string does enable adding any path or URI that contains the colon, but at the cost of having to escape using the `\\`.

If we add a negative lookahead for `//`, assuming that normal paths would never start with `//`, then we can support http and ftp URIs without escaping:

```ruby
ENV.fetch("OOD_QUOTA_PATH", "").strip.split(/(?<!\\):(?!\/\/)/).map{ |p| p.gsub(/\\/,"") }
```

For example

```
irb(main):008:0> path = "/project/rice/d/ddietz:https://www.rcac.purdue.edu/ws/storagedirquota/ddietz:/scratch/rice/d/ddietz:/path/with/\\:colon/in/it"
=> "/project/rice/d/ddietz:https://www.rcac.purdue.edu/ws/storagedirquota/ddietz:/scratch/rice/d/ddietz:/path/with/\\:colon/in/it"
irb(main):016:0> path.split(/(?<!\\):(?!\/\/)/).map{ |p| p.gsub(/\\/,"") }
=> ["/project/rice/d/ddietz", "https://www.rcac.purdue.edu/ws/storagedirquota/ddietz", "/scratch/rice/d/ddietz", "/path/with/:colon/in/it"]
```

Another option is to drop support for escaping the colon, since the real goal right now is to only add support for URLs. And in that case we can do:

```ruby
def quota_paths
  ENV.fetch("OOD_QUOTA_PATH", "").strip.split(/:(?!\/\/)/)
end
```

then:

```
irb(main):018:0> path = "/project/rice/d/ddietz:https://www.rcac.purdue.edu/ws/storagedirquota/ddietz:/scratch/rice/d/ddietz"
=> "/project/rice/d/ddietz:https://www.rcac.purdue.edu/ws/storagedirquota/ddietz:/scratch/rice/d/ddietz"
irb(main):019:0> path.split(/:(?!\/\/)/)
=> ["/project/rice/d/ddietz", "https://www.rcac.purdue.edu/ws/storagedirquota/ddietz", "/scratch/rice/d/ddietz"]
```

IMHO this would be my preference since the code is a little simpler and we don't have a use case for escaping the colon beyond supporting URLs. @MorganRodgers any thoughts?

MorganRodgers: I prefer the second as it is simpler but still meets our needs. Symbol heavy regex take longer to grok.

ericfranz: @ddietz89 I just wanted to follow up and confirm if you were okay with the proposed change, where instead of supporting escaping colon in the OOD_QUOTA_PATH, we instead would use negative lookahead for `//` to ignore any colon that preceeded `//` which would enable support for all URLs of the format `scheme://`

ddietz89: @ericfranz Sorry, have been rolling out OOD site wide and forgot about this. Yes, this makes sense. I switched to this method in my deployed version of the code. Thanks!

ericfranz: Done! https://github.com/OSC/ood-dashboard/commit/f41ff2ea9f0e9f28b34055d522758f2fec1b4cbb

PR #456, opened 2019-03-19
[EXAMPLE] - Example of altered vnc.html

MorganRodgers:
The (re)connect button is mostly (entirely?) seen by users whose VNC session has ended. As such the ability to reconnect to the session is not actually offered by this control and its utility is limited. This example vnc.html and base.css remove the (re)connect button in favor of a message informing the user that their session has ended, and it is now safe to close the window.

At this time this is not intended for inclusion into OnDemand's master branch.

### DIFF
--- a/public/noVNC-1.0.0/app/styles/base.css
+++ b/public/noVNC-1.0.0/app/styles/base.css
@@ -900,3 +900,30 @@ select:active {
     font-size: 90px;
   }
 }
+
+/* ----------------------------------------
+ * Site Specific Customizations
+ * ----------------------------------------
+ */
+
+#safe_to_close_notification div {
+  margin: 2px;
+  padding: 5px 30px;
+  border: 1px solid rgb(83, 99, 122);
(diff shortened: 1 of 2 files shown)


MorganRodgers:
What I don't like about the minimalist JS solution is the JS event listener that attempts to reconnect to the VNC  session is still installed on the element. Without having a handle to the event listener, it appears that we would need to replace the element with its clone which sounds pleasingly science-fictiony: https://stackoverflow.com/a/4386514/1015786.

ericfranz:
@MorganRodgers which event listener. Can you point to the code?

MorganRodgers:
The event listener `noVNC_connect_button` is set: https://github.com/OSC/ood-dashboard/blob/master/public/noVNC-1.0.0/app.js#L480-L486

The same code is also present in https://github.com/OSC/ood-dashboard/blob/master/public/noVNC-1.0.0/app/ui.js

PR #458, opened 2019-03-26
Namespace translations under generic or site specific and flatten hierarchies

MorganRodgers:
There are still un-translated items on the Sandbox Apps page: `/pun/dev/ood-dashboard/admin/dev/products`.

Once translations are complete 👍 needs to be removed from the translation file, and the old commented-out translations deleted. Note that the thumbs are why the CI checks failed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,6 @@ tags
 # Ignore data files
 /data
 config/clusters.d
+
+# SSHFS
+._*
--- a/app/controllers/batch_connect/session_contexts_controller.rb
+++ b/app/controllers/batch_connect/session_contexts_controller.rb
@@ -33,7 +33,7 @@ class BatchConnect::SessionContextsController < ApplicationController
     respond_to do |format|
       if @session.save(app: @app, context: @session_context, format: @render_format)
         cache_file.write(@session_context.to_json)  # save context to cache file
-        format.html { redirect_to batch_connect_sessions_url, notice: t('dashboard.generic_msg.batch_connect_sessions_status_blurb_create_success') }
+        format.html { redirect_to batch_connect_sessions_url, notice: t('dashboard.tr.batch_connect_sessions_status_blurb_create_success') }
         format.json { head :no_content }
       else
         format.html do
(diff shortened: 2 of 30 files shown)


ericfranz:
Proposal:

Change `site_specific` to `site` and `generic_msg` to `tr`. Reason: `tr` in internationalization is a common abbreviation form of "translation" (even the tr Unix/Linux command is for translate). And that's the purpose of those strings.

With so many translations, and only two scopes (site and tr), I think this might be an exception to the normal rule of preferring longer full words for the scopes like we would for any other variable name. In this case I think it would improve readability slightly because everywhere you look the bulk of the string is the app name and the string name i.e. compare:

- **dashboard**.generic_msg.**batch_connect_sessions_status_blurb_create_success**
- **dashboard**.tr.**batch_connect_sessions_status_blurb_create_success**

Of course if there was only one or two cases, the longer might be preferable, though "translation" might be preferable to "generic_msg". But since this is everywhere...

ericfranz:
This looks great. Unfortunately I just realized that the existing OnDemand supports these two:

https://github.com/OSC/ood-dashboard/blob/f41ff2ea9f0e9f28b34055d522758f2fec1b4cbb/config/locales/en.yml#L22-L27

So if we move these to site i.e.

```yaml
en:
  dashboard:
    site:
      welcome_html:
      motd_title:
    tr:
      ...
```

we should at least support backwards compatibility with existing overrides that specify values for  `en.dashboard.welcome_html` and `en.dashboard.motd_title`.

Another option is we omit "site" and unindent the site specific strings. This would then be backwards compatible:

```yaml
en:
  dashboard:
    welcome_html:
    motd_title:
    batch_connect_sessions_status_bad: "Your session has entered a bad state. Feel free to contact support for
            further information."
    batch_connect_sessions_status_missing_connection: |
        This app is missing information required to establish a connection. Please
        contact support if you see this message.
    batch_connect_sessions_status_queued: |
        Please be patient as your job currently sits in queue. The wait time depends
        on the number of cores as well as time requested.
    nav_help_change_password: "Change HPC Password"
    nav_help_docs: "Online Documentation"
    nav_help_support: "Contact Support"
    nav_help_two_factor: "Configure Two-Factor Authentication"
    quota_additional_message: "Consider deleting or archiving files to free up disk space."
    quota_reload_message: "Reload page to see updated quota. Quotas are updated every 5 minutes."
    quota_warning_prefix_html: "Quota limit warning for"
    tr:
      apps_setup_failure_html:
      apps_system_apps_title: "System Apps"
      batch_connect_form_data_root: "data root directory"
      ...
```

While this might not be the purest organization, it might be the easiest to supporting the preexisting keys and still accomplish the goal of grouping the keys we want to call attention to at the top and the rest in the tr group below.

MorganRodgers:
Stability in products is a virtue; agreed. I will remove the `site` namespace in the translation for now.

PR #460, opened 2019-03-26
Display VERSION string for interactive app in web form

MorganRodgers:
Prefers `VERSION` files, but falls back to `git describe --always --tags` otherwise.

### DIFF
--- a/app/models/batch_connect/app.rb
+++ b/app/models/batch_connect/app.rb
@@ -202,6 +202,17 @@ module BatchConnect
       @sub_app_list ||= build_sub_app_list
     end
 
+    # The version of the OodApp
+    # @return [String] the version
+    def version
+      @ood_app.version
+    end
+
+    # @return [String] the app name
+    def name
+      @ood_app.manifest.name.empty? ? @ood_app.name : @ood_app.manifest.name
+    end
+
     # Convert object to string
     # @return [String] the string describing this object
     def to_s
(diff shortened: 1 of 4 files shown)


app/models/batch_connect/app.rb
```diff
@@ -202,6 +202,17 @@ def sub_app_list
...
+    # The version of the OodApp
+    # @return [String] the version
+    def version
+      @ood_app.version
+    end
+
+    # @return [String] the app name
+    def name
+      @ood_app.manifest.name.empty? ? @ood_app.name : @ood_app.manifest.name
+    end
```
ericfranz: What are you trying to achieve by adding this method `name` as opposed to using `title`?

Both `OodApp` and `BatchConnect::App` implement `title`. In `OodApp` the `title` is the human readable title that the manifest can provide and the `name` is the directory name and what appears in the URL.

The weird thing with BatchConnect apps is their title is set in the form.yml, which is something that will be "fixed" in another issue, which will attempt to avoid rendering erbs when building the menu.

Whatever you do you will have the case where BatchConnect "subapps" i.e. the desktop configs.

Also I would rather see the version in a more prominent location. Perhaps in the title using `<small>` since you have to scroll to see the footer otherwise. With this HTML as an example `<h3>Owens Desktop <small>version: 1.2.6</small></h3>` you end up with this:

screen 2019-03-26 at 9 23 25 PM

Finally I know I suggested the default should display "unknown" but the default should probably be hidden in this case, because sites will have custom interactive apps, possibly without .git dir or VERSION file in the root of those directories, and we wouldn't want this new "version: unknown" to display.

PR #461, opened 2019-03-27
Fix order of bc save for better error handling

### LINKED ISSUE #451: Fix order of interactive app submission handling
This code handles the form submission for an interactive app in this order:

1. stages the app template, providing the context for rendering the template files
2. renders the submit.yml.erb file providing the context to get a submit hash
3. calls submit with the with the hash to submit a job

https://github.com/OSC/ood-dashboard/blob/baea5558a4e55e8f7ca49599670e6731ad2da94b/app/models/batch_connect/session.rb#L137-L147

The rendering of the submit.yml.erb should occur first. That way, if there are any errors, the submit.yml.erb can raise an exception and the error will be presented to the user. Currently the user receives the error but after the template is copied to the user's home directory.

So the order would instead be 

1. renders the submit.yml.erb file providing the context to get a submit hash
2. stages the app template, providing the context for rendering the template files
3. calls submit with the with the hash to submit a job

```ruby
def save(app:, context:, format: nil)
  # ...
  submit_script = app.submit_opts(context, fmt: format) # raises an exception
  stage(app.root.join("template"), context: context) && submit(submit_script)

rescue => e   # rescue from all standard exceptions (app never crashes)
  errors.add(:stage, e.message)
  Rails.logger.error("ERROR: #{e.class} - #{e.message}")
  false
end
```

ericfranz:
Fixes #451

when saving batch connect session:

build script from form submission and submit.yml prior to staging
that way if the script creation raises an exception it is handled
without staging files

### DIFF
--- a/app/models/batch_connect/session.rb
+++ b/app/models/batch_connect/session.rb
@@ -142,8 +142,14 @@ module BatchConnect
       self.view       = app.session_view
       self.created_at = Time.now.to_i
 
+      submit_script = app.submit_opts(context, fmt: format) # could raise an exception
+
       stage(app.root.join("template"), context: context) &&
-        submit(app.submit_opts(context, fmt: format))
+        submit(submit_script)
+    rescue => e   # rescue from all standard exceptions (app never crashes)
+      errors.add(:save, e.message)
+      Rails.logger.error("ERROR: #{e.class} - #{e.message}")
+      false
     end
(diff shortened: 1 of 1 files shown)


ericfranz:
Test and verify you can raise errors in submit.yml.erb and not result in any files being staged.

Issue #441, opened 2019-01-09
Lazy exception handling for announcement files pollutes log file


ericfranz:
1. If an announcement file doesn't exist (which we test multiple ones) we log a "warning" for each: https://github.com/OSC/ood-dashboard/blob/f0ba96fb4c800348293f3ce1f2db7f2e1468b678/app/models/announcement.rb#L38-L51
2. reading a single announcement raises an un-handled exception, no announcements are displayed https://github.com/OSC/ood-dashboard/blob/f0ba96fb4c800348293f3ce1f2db7f2e1468b678/app/controllers/application_controller.rb#L49-L54
3. we rescue from StandardError, which is too broad

ericfranz:
In regards to how announcements work... (and the related RSS announcements), we might expect this to work like quota files work - you can add URLs to the path string for announcements.

We could still have URLs processed client side instead of server side if we want. In that case we could cherry pick from the RSS implementation and use that if the response is XML. If the response is HTML (or text) it could just be embedded in the alert itself.

PR #462, opened 2019-03-29
Handle file not found error for announcements

MorganRodgers:
Stop showing the parse error warning in the logs when no announcement files exist by handling `Errno::ENOENT`.

### DIFF
--- a/app/models/announcement.rb
+++ b/app/models/announcement.rb
@@ -45,6 +45,8 @@ class Announcement
                   {}
                 end
       @opts.symbolize_keys.compact
+    rescue Errno::ENOENT  # File does not exist
+      @opts = {}
     rescue => e
       Rails.logger.warn "Error parsing announcement file '#{@path}': #{e.message}"
       @opts = {}


MorganRodgers:
Fixes #441.

@ericfranz FYI.

ericfranz:
This class should be rewritten. There is this "opts" that means different things throughout the app.

For example, it appears that the app can be instantiated using a pathname or a string that is a path, or a hash form of the file that would be loaded from reading the the file and the path location. Too much happening in the initializer. Better to have a factory method like `Announcement.from_file(path)` and have the initializer to use the attribute keys `Announcement.new(msg:, type:)`. Like:

```ruby
class Announcement
  include ActiveModel::Model
  attr_accessor :msg, :type
  validates :msg, :type, presence: true

  def self.from_file(path)
    # ...
  end
 end
```

Is the lazy reading of the file done to avoid reading the file contents if the announcement data will not be used?

ericfranz:
Lets save the refactor for when we touch this again later.

PR #467, opened 2019-04-19
[WIP] - Enable site wite option to open apps in same tab instead of new tab

### LINKED ISSUE #466: Enable site wide option to open apps in same window instead of new window
Easier than "user configurable" and might be enough to satisfy some of our external partners.

Two values:

1. `OOD_DASHBOARD_OPEN_EXTERNAL_LINKS_NEW_WINDOW=1` or truthy => external links and apps open in new tab - this is the current functionality
3. `OOD_DASHBOARD_OPEN_LINKS_NEW_WINDOW=0` or false or just not set at all => open apps in same window - we can make this the new default

Then everywhere we set target="_blank" in the Dashboard we can instead use target="<%= Configuration.external_link_target %>" where that helper returns something like:

```ruby
# in ConfigurationSingleton

def external_link_target
  @external_link_target ||= (to_bool(ENV['OOD_DASHBOARD_OPEN_EXTERNAL_LINKS_NEW_WINDOW') ? "_blank" : "_self")
end
```

In a trivial test though I did have links that had target="_self" opening in a new window so that was confusing to me. Please test!

MorganRodgers:
Fixes  #466.

### DIFF
--- a/app/apps/ood_app.rb
+++ b/app/apps/ood_app.rb
@@ -68,8 +68,7 @@ class OodApp
           description: manifest.description,
           url: OodAppkit::Urls::Files.new(base_url: url).url(path: Dir.home),
           icon_uri: "fas://home",
-          caption: caption,
-          new_tab: true
+          caption: caption
         )
       ].concat(
         OodFilesApp.new.favorite_paths.map do |path|
@@ -78,8 +77,7 @@ class OodApp
             description: manifest.description,
             url: OodAppkit::Urls::Files.new(base_url: url).url(path: path),
             icon_uri: "fas://folder",
-            caption: caption,
-            new_tab: true
+            caption: caption
           )
         end
       )
@@ -97,8 +95,7 @@ class OodApp
             description: manifest.description,
             url: OodAppkit::Urls::Shell.new(base_url: url).url,
             icon_uri: "fas://terminal",
-            caption: caption,
-            new_tab: true
+            caption: caption
           )
         ]
       else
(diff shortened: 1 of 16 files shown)


app/apps/ood_app.rb
```diff
@@ -98,7 +98,7 @@ def links
             url: OodAppkit::Urls::Shell.new(base_url: url).url,
             icon_uri: "fas://terminal",
             caption: caption,
-            new_tab: true
+            new_tab: Configuration.open_link_in_new_window?
```
ericfranz: As discussed, force open shell in new tab

app/models/batch_connect/app.rb
```diff
@@ -110,7 +110,7 @@ def link
         url: Rails.application.routes.url_helpers.new_batch_connect_session_context_path(token: token),
         icon_uri: ood_app.icon_uri,
         caption: ood_app.caption,
-        new_tab: false
+        new_tab: Configuration.open_link_in_new_window?
```
ericfranz: As discussed, force batch connect always to be in same window (false new tab)

ericfranz:
Other thought: `external_link_target` might be a misnomer here, as it is being used for the targets for both internal and external links, where internal might be the Files app or the Job Composer and external might be Contact Support.

ericfranz:
A: Preexisting behavior: External to Dashboard opens in new window

- New Window: Shell, NoVNC/Jupyter connect, Files, Job Composer, Active Jobs, Help links, Dev Documentation
- Same Window: Interactive Apps, Interactive Sessions, Develop

B: 

- New Window: Shell, NoVNC/Jupyter connect
- Same Window: every other link

C:

- No new windows

D: 

- New window for links external to ondemand.osc.edu

PR #468, opened 2019-04-24
Refactor Quota code

### LINKED ISSUE #453: Refactor code that manages reading the quota files
The code for parsing quota files and generating an array of Quota objects is complex, with most of it split between several class methods on Quota. Complexity has now increased due to adding the ability to load from a URL instead of a Path.

It may be appropriate to add a new QuotaFile class to encapsulate this complexity. That way some of the logic in the static methods can be spread across a few more instance methods. This would make it easier to group automated tests for the QuotaFile parsing itself. If we had a QuotaFile class we might consider moving the parsing of the colon delimited paths and urls to this class as well as a factory method. `ConfigurationSingleton#quota_paths` could even be updated to return an array of `QuotaFile` objects.

Also, since we are now working with file paths or URLs, consider investigating using https://ruby-doc.org/stdlib-2.4.0/libdoc/open-uri/rdoc/OpenURI.html so you can do something like:

```ruby
open(quota_uri).read
```

to get the contents of the JSON file, whether quota_uri is a system path or a URL. Note I don't know the ramifications of using`open-uri` vs `net/http` for HTTP requests, especially in terms of error handling, which is especially important with HTTP requests.

ericfranz:
Fixes #453

### DIFF
--- a/app/models/quota.rb
+++ b/app/models/quota.rb
@@ -18,28 +18,16 @@ class Quota
     # KeyError and JSON::ParserErrors shall be non-fatal errors
     def find(quota_path, user)
       user  = user && user.to_s
-
-      quotas = []
-
       raw = open(quota_path).read
       return [] if raw.nil?
 
       # Attempt to parse raw JSON into an object
       json = JSON.parse(raw)
 
-      # Parse JSON object into quota data
-      begin
-        case json["version"].to_i
-        when 1
-          quotas += find_v1(user, json)
-        else
-          raise InvalidQuotaFile.new("JSON version found was: #{json["version"].to_i}")
-        end
-      rescue KeyError => e
-        Rails.logger.error("Quota entry for user #{user} is missing expected parameter #{e.message}")
-      end
-
-      quotas
+      #FIXME: any validation of the structure here? otherwise we don't need the complexity of the code below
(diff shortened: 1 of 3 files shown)


ericfranz:
Is this more readable:

```ruby
def build_quotas(quota_hashes, updated_at, user)
  quota_hashes.select { |quota|
    user == quota["user"]
  }.reduce([]) { |quotas, quota|
    quotas += create_both_quota_types(quota.merge("updated_at" => quota.fetch("timestamp", updated_at)))
  }
end
```

or this


```ruby
def build_quotas(quota_hashes, updated_at, user)
  quota_hashes.reduce([]) do |quotas, quota|
    quotas += create_both_quota_types(quota.merge("updated_at" => quota.fetch("timestamp", updated_at))) if user == quota["user"]
    quotas
  end
end
```
than this

```ruby
def build_quotas(quota_hashes, updated_at, user)
  q = []
  quota_hashes.each do |quota|
    q += create_both_quota_types(quota.merge("updated_at" => quota.fetch("timestamp", updated_at))) if user == quota["user"]
  end
  q
end
```

MorganRodgers:
I like the first with the select, but TBH the third is probably what I would have written to start with.

PR #474, opened 2019-05-28
Quality Setting Manipulation in NoVNC

samirmansour:
I have added the query strings to manipulate the compression level and the quality level. If no compression and quality levels are provided in the query it just uses the defaults.

### DIFF
--- a/public/noVNC-1.0.0/app.js
+++ b/public/noVNC-1.0.0/app.js
@@ -7497,6 +7497,20 @@ RFB.prototype = {
     _sendEncodings: function () {
         var encs = [];
 
+        //URLSearchParams parses through the query string in search bar
+        var urlParams = new URLSearchParams(window.location.search);
+
+        var qualityLevelDefault = _encodings.encodings.psuedoEncodingQualityLevel0 + 6;
+        
+        var compressionLevelDefault = _encodings.encodings.psuedoEncodingCompressLevel0 + 2;
+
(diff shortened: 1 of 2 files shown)


public/noVNC-1.0.0/app.js
```diff
@@ -7497,6 +7497,20 @@ RFB.prototype = {
     _sendEncodings: function () {
         var encs = [];
 
+        //URLSearchParams parses through the query string in search bar
+        var urlParams = new URLSearchParams(window.location.search);
+
+        var qualityLevelDefault = _encodings.encodings.psuedoEncodingQualityLevel0 + 6;
+        
+        var compressionLevelDefault = _encodings.encodings.psuedoEncodingCompressLevel0 + 2;
```
ericfranz: default values are misspelled

it is pseudo not psuedo

ericfranz:
@samirmansour take a look at the comment I added.

Also can you confirm whether or not changes to the settings have effect every time you click connect? i.e.

1. after a session starts you can change the setting, click connect, verify the settings effect, close the novnc browser tab
2. change the setting to something else, click connect, and verify changes had an impact

ericfranz:
Also I would add `target: "_blank"` to the connect button so you can open it in a new window by default

ericfranz:
Actually I did a quick test and it appears that the settings are applied every time you connect!

PR #476, opened 2019-06-05
Add VNC setting controls to noVNC launcher

MorganRodgers:


MorganRodgers:
Replaces #474.

MorganRodgers:
@ericfranz I like where this has ended up. Shall we aim to include this in 1.6?

PR #483, opened 2019-08-30
Support balances and add warnings

treydock:
Screenshot using default locales

Screen Shot 2019-08-30 at 3 15 49 PM

Screenshot using OSC locales

Screen Shot 2019-08-30 at 3 13 36 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,7 +219,7 @@ Where `version` defines the version of the JSON schema used, `timestamp`
 defines when this file was generated, and `balances` is a list of balance objects
 (see below).
 
-The value for `config.unit` defines the type of units for balances and `config.project_type` would be project, account, or group, etc.
+The value for `config.unit` defines the type of units for balances and `config.project_type` would be project, account, or group, etc. Both values are used in locales and can be any string value.
 
 You can configure the Dashboard to use this JSON file (or files) by setting the
 environment variable `OOD_BALANCE_PATH` as a colon-delimited list of all JSON
--- a/README.md
+++ b/README.md
@@ -189,6 +189,77 @@ be repeated with just `user`, `block_usage`, and `file_usage` changing.*
 
 *Warning: A block must be equal to 1 KB for proper conversions.*
 
+### Balance Warnings
+
+We currently support displaying warnings to users on the Dashboard if their
+balance is nearing its limit. This requires an auto-updated (it is
+recommended to update this file every **15 minutes** with a cronjob) JSON file
+that lists all user balances. The JSON schema for version `1` is given as:
+
+```json
+{
+  "version": 1,
+  "timestamp": 1525361263,
+  "config": {
+    "unit": "RU",
+    "project_type": "project"
+  },
+  "balances": [
+    {
+      ...
+    },
+    {
+      ...
+    }
+  ]
+}
+```
+
+Where `version` defines the version of the JSON schema used, `timestamp`
+defines when this file was generated, and `quotas` is a list of quota objects
+(see below).
+
+The value for `config.unit` defines the type of units for balances and `config.project_type` would be project, account, or group, etc.
+
+You can configure the Dashboard to use this JSON file (or files) by setting the
+environment variable `OOD_BALANCE_PATH` as a colon-delimited list of all JSON
+file paths.
+
+The default threshold for displaying the warning is at `0`, but this
+can be changed with the environment variable `OOD_BALANCE_THRESHOLD`.
+
+An example is given as:
+
+```shell
+# /etc/ood/config/apps/dashboard/env
+
+OOD_BALANCE_PATH="/path/to/balance1.json:/path/to/balance2.json"
+OOD_BALANCE_THRESHOLD=1000
+```
+
+#### User Balance
+
+If the balance is defined as a `user` quota, then it applies to only that user. Omit the `project` key:
+
+```json
+{
+  "user": "user1",
+  "value": 10
+}
+```
+
+#### Project Balance
+
+If the balance is defined as a `project` balance, then it applies to a project/account/group, whatever is defined for `config.project_type`:
+
+```json
+{
+  "user": "user1",
+  "project": "project1",
+  "value": 10
+}
+```
+
 ## API
 
 ### iHPC CLI
--- a/README.md
+++ b/README.md
@@ -216,7 +216,7 @@ that lists all user balances. The JSON schema for version `1` is given as:
 ```
 
 Where `version` defines the version of the JSON schema used, `timestamp`
-defines when this file was generated, and `quotas` is a list of quota objects
+defines when this file was generated, and `balances` is a list of balance objects
 (see below).
 
 The value for `config.unit` defines the type of units for balances and `config.project_type` would be project, account, or group, etc.
--- a/README.md
+++ b/README.md
@@ -193,7 +193,7 @@ be repeated with just `user`, `block_usage`, and `file_usage` changing.*
 
 We currently support displaying warnings to users on the Dashboard if their
 balance is nearing its limit. This requires an auto-updated (it is
-recommended to update this file every **15 minutes** with a cronjob) JSON file
+recommended to update this file every day with a cronjob) JSON file
 that lists all user balances. The JSON schema for version `1` is given as:
 
 ```json
--- a/app/models/balance.rb
+++ b/app/models/balance.rb
@@ -94,6 +94,6 @@ class Balance
   end
 
   def to_s
-    I18n.translate('dashboard.balance_message', units_balance: units_balance, value: @value)
+    I18n.translate('dashboard.balance_message', unit: @unit, units_balance: units_balance, value: @value)
   end
 end
--- a/app/models/balance.rb
+++ b/app/models/balance.rb
@@ -75,13 +75,13 @@ class Balance
 
   # Unit + balance text
   def units_balance
-    "#{@unit} balance" if @unit.present?
+    return "#{@unit} balance" if @unit.present?
     'Balance'
   end
 
   # Plural units
   def balanace_units
-    @unit.pluralize if @unit.present?
+    return @unit.pluralize if @unit.present?
     'resources'
   end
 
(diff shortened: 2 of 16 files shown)


app/models/balance.rb
```diff
@@ -0,0 +1,87 @@
+# This describes disk quota utilization for a given user and volume
```
ericfranz: Update comment to describe balance instead of disk quota. Maybe in parentheses (RUs) in comments for us on Gateways?

app/models/balance.rb
```diff
@@ -0,0 +1,87 @@
...
+    @project = params.fetch(:project, nil).to_s
+    @value = params.fetch(:value).to_i
+    @project_type = params.fetch(:project_type, nil).to_s
+    @unit = params.fetch(:unit).to_s
+    @updated_at = Time.at(params.fetch(:updated_at).to_i)
+  end
+
+  def balance_object
+    return @project if @project.present?
+    @user
```
ericfranz: Can use the new [presence](https://guides.rubyonrails.org/active_support_core_extensions.html#presence) here:

```ruby
def balance_object
  project.presence || user
end
```

config/configuration_singleton.rb
```diff
@@ -109,6 +109,25 @@ def quota_threshold
...
+  # @return [Array<String>] balance paths
+  def balance_paths
+    # regex uses negative lookahead to ignore : preceeding //
+    ENV.fetch("OOD_BALANCE_PATH", "").strip.split(/:(?!\/\/)/)
+  end
+
+  # The threshold for determining if there is sufficient balance remaining
+  # @return [Float] threshold factor
+  def balance_threshold
+    ENV.fetch("OOD_BALANCE_THRESHOLD", 0).to_f
```
ericfranz: The default threshold should be higher than 0. If 0, the warning message will always display, if this works like the quota warnings do.

README.md
```diff
@@ -189,6 +189,77 @@ be repeated with just `user`, `block_usage`, and `file_usage` changing.*
...
+    },
+    {
+      ...
+    }
+  ]
+}
+```
+
+Where `version` defines the version of the JSON schema used, `timestamp`
+defines when this file was generated, and `quotas` is a list of quota objects
```
ericfranz: you mean `balances` not `quotas`

README.md
```diff
@@ -189,6 +189,77 @@ be repeated with just `user`, `block_usage`, and `file_usage` changing.*
...
+(see below).
+
+The value for `config.unit` defines the type of units for balances and `config.project_type` would be project, account, or group, etc.
+
+You can configure the Dashboard to use this JSON file (or files) by setting the
+environment variable `OOD_BALANCE_PATH` as a colon-delimited list of all JSON
+file paths.
+
+The default threshold for displaying the warning is at `0`, but this
```
ericfranz: Should the default be something other than 0?

ericfranz: Oh I see threshold is not a percentage but an integer value like 1000.

config/configuration_singleton.rb
ericfranz: Sorry I realize now that the threshold value is an integer i.e. number of RUs, rather than a percentage. Markign this as resolved.

README.md
```diff
@@ -189,6 +189,77 @@ be repeated with just `user`, `block_usage`, and `file_usage` changing.*
...
+    }
+  ]
+}
+```
+
+Where `version` defines the version of the JSON schema used, `timestamp`
+defines when this file was generated, and `quotas` is a list of quota objects
+(see below).
+
+The value for `config.unit` defines the type of units for balances and `config.project_type` would be project, account, or group, etc.
```
MorganRodgers: You should explicitly state that this value is used in the locale files, and so can be any string value that makes sense.

app/models/balance.rb
MorganRodgers: I would prefer that the comment refer to `config.unit` specifically, and call out resource units as what are used at OSC.

treydock: Shouldn't that go in README or does it belong in model header too?

app/models/balance.rb
treydock: Change made.

README.md
treydock: Fixed copy+paste mistake

README.md
treydock: Updated to call both out.

treydock:
I updated it so `config` key is optional and added better default locale text if no units are provided

Screen Shot 2019-09-03 at 1 54 26 PM

treydock:
Slightly better:

Screen Shot 2019-09-03 at 2 01 25 PM

treydock:
Turns out updating every 15 minutes is useless because OSC only sends charging data to database once a day, which seems logical other sites do things similarly as easier to charge for previous day than several times a day.  Going to remove the refresh locale.

treydock:
Actually might leave text but modify for default to daily.

treydock:
One thing I added and can revert if not wanted is concept of OnDemand having timezone configured. The ActiveSupport time zone names I couldn't get from local server timezone via `Time` so had to configure the Rails app. I couldn't set time zone with initializer since has to be set in `Application` class. The absence of a timezone just sticks with UTC which is the default. Since the time used for things like balances is specific to where the server is deployed and not the user's timzone, I opted for server-side time zone configuration.

treydock:
This is current look using OSC specific units and project type but default locales.

Screen Shot 2019-09-04 at 8 04 14 AM

config/configuration_singleton.rb
```diff
@@ -109,6 +109,25 @@ def quota_threshold
...
+  # Can be URL or File path. colon delimited string; though colon in URL is
+  # ignored if URL has format: scheme://path (colons preceeding // are ignored)
+  #
+  # /path/to/balance.json:https://osc.edu/balance.json
+  #
+  #
+  # @return [Array<String>] balance paths
+  def balance_paths
+    # regex uses negative lookahead to ignore : preceeding //
+    ENV.fetch("OOD_BALANCE_PATH", "").strip.split(/:(?!\/\/)/)
```
ericfranz: we should probably memoize this and `quota_paths` since this is executed on every request...

```ruby
def balance_paths
  @balance_paths ||= ENV.fetch("OOD_BALANCE_PATH", "").strip.split(/:(?!\/\/)/)
end
```

treydock: I made the change but is looking up something in the environment and doing a few operations on the value really that expensive?

ericfranz: No its not. But its also easy to memoize. I guess we didn't do that before because we didn't actively memoize everything we could. Maybe its a bad idea in practice to do that there and instead utilize view caching where appropriate. For example, parsing quota files is far more expensive than getting the list of files to parse by using a regex.

ericfranz:
We should actually revert the timezone change. In the Job Composer we use this Rails gem https://github.com/basecamp/local_time. That adds a view helper "local_time" so you can do something [like this](https://github.com/OSC/ood-myjobs/blob/41cb76814f28387085e419eebe78ea472ed1d99d/app/views/workflows/index.html.erb#L84):

```
<%= local_time(workflow.created_at) %>
```

The result is the user sees something in their own time zone instead of UTC time formatted string. Thus, if I create a new job at Utah's CHPC or at OSC in the Job Composer the created at displays like this: "September 6, 2019 2:35pm"  - even though the time zones are different I know as a user I created both of those 15 minutes ago since it is 2:50 PM right now.

If you want to display this string in local time that would be the most appropriate to do now.

As far as having an OOD_TIMEZONE, you could open a separate issue on the ondemand repo about the topic of displaying time in OnDemand and we could adopt a more formalized strategy that addresses drawbacks or insufficiency of using Basecamp's local_time gem.

config/configuration_singleton.rb
ericfranz: You know what adding that memoization was probably bad advice. I vaguely remember avoiding or removing excessive memoization in this file for this reason. If you want to revert you can.

treydock:
Added commit that uses `local_time` to generate the last_update value.

app/views/shared/_insufficient_balance_resource.html.erb
```diff
@@ -2,7 +2,7 @@
   <%= t('dashboard.balance_warning_prefix_html', units_balance: balance.units_balance, unit: @unit) %>
   <strong><%= balance.project_type %> <%= balance.balance_object %></strong>
   <small class="pull-right hidden-xs hidden-sm">
-    <%= t('dashboard.balance_reload_message_html', last_update: balance.last_update) %>
+    <%= t('dashboard.balance_reload_message_html', last_update: local_time(balance.last_update)) %>
```
ericfranz: This could just be:

```
    <%= t('dashboard.balance_reload_message_html', last_update: local_time(balance.updated_at)) %>
```

and `Balance#last_update` removed as `Balance#last_update` purpose is to make a local time human readable form of `Balance#updated_at`, which is what `local_time(balance.updated_at)` also achieves.

app/views/shared/_insufficient_balance_resource.html.erb
```diff
@@ -0,0 +1,12 @@
+<h4>
+  <%= t('dashboard.balance_warning_prefix_html', units_balance: balance.units_balance, unit: @unit) %>
```
ericfranz: `@unit` is an instance variable of `Balance` so should this instead be:

```
  <%= t('dashboard.balance_warning_prefix_html', units_balance: balance.units_balance, unit: balance.unit) %>
```

app/views/shared/_insufficient_balance_resource.html.erb
```diff
@@ -0,0 +1,12 @@
...
+  <strong><%= balance.project_type %> <%= balance.balance_object %></strong>
+  <small class="pull-right hidden-xs hidden-sm">
+    <%= t('dashboard.balance_reload_message_html', last_update: local_time(balance.last_update)) %>
+  </small>
+</h4>
+<ul class="list-unstyled">
+  <li>
+    <%= balance.to_s %>. <%= t('dashboard.balance_additional_message', balanace_units: balance.balanace_units, unit: @unit) %>
```
ericfranz: See comment above, it seems like `@unit` here should actually be `balance.unit`

